PR #125, opened 2023-10-31
Add metadata for OpenAPI and JsonSchema

spinillos:
It reads metadata for OpenAPI and JsonSchema. 

The metadata is extracted inside `GenerateAST` from the schema to avoid to have the option to set this information in the configuration too.


K-Phoen:
@spinillos should we close this PR and implement this again if/when needed?

Issue #159, opened 2023-11-21
Fields with struct references and defaults aren't well parsed in cue.


spinillos:
Use case: https://github.com/grafana/kind-registry/blob/main/grafana/next/composable/candlestick/panelcfg.cue#L48

```cue
colors: CandlestickColors | *{
    down: "red"
    up:   "green"
    flat: "gray"
}
```

Generates:
GO
```go
Colors struct {
    Down string `json:"down"`
    Up   string `json:"up"`
    Flat string `json:"flat"`
}
```

TS:
```ts
colors: {
    down: "red";
    up: "green";
    flat: "gray";
};
```

When it should set `CandlestickColors` struct and the values as defaults.

PR #175, opened 2023-11-24
Parse struct defaults in cue

spinillos:
Closes https://github.com/grafana/cog/issues/159

It parses cue struct defaults as `map[string]interface{}`, mainly to be able to test it. 
In Go, it sets the correspondent builder since in TS it only set its values for the struct. I found some import issues using the builders in TS and its why it sets only the values.

Edit: The PR became a huge rabbit hole and I'm sorry for that 😞. This PR covers:

* It allows defined struct with defaults with nested values and/or structs.
* It allows anonymous structs with defaults with nested values and/or structs.
* For TS, when a struct has a default, all of its fields should be filled, but cue allows to set them partially. So it fills the missing ones with default values. [[code]](https://github.com/grafana/cog/pull/175/files#diff-2b901cf064fedf690f5ef531d71c56b61dc01f95077dcd614ddd512908dd423bR331-R339)
* For Go, when the default is a reference and it has a builder, it generates the default using the builder instead of a struct. [[code]](https://github.com/grafana/cog/pull/175/files#diff-030991de1bcfed9baf3358c262ade44e60a95594442442208741c86ea96d3538R157-R162)
* It adds a `tools` jenny for struct defaults that contains pointers, to transform the values ☠️ [[code]](https://github.com/grafana/cog/pull/175/files#diff-4fa5f521637a17ff088fbf2758447735b16c59c62af015988b3c59145b2fe412R191-R193)
* There are some `TODO`, `FIXME` comments to map some missing use cases that we could find in a future and/or some checks that "shouldn't" be necessary if the use case appears.

### DIFF
--- a/internal/jennies/golang/jennies.go
+++ b/internal/jennies/golang/jennies.go
@@ -68,7 +68,6 @@ func (language *Language) Jennies(globalConfig common.Config) *codejen.JennyList
 
 		common.If[common.Context](globalConfig.Builders, &Builder{Config: config}),
 	)
-	jenny.AppendOneToOne(Tools{})
 	jenny.AddPostprocessors(PostProcessFile, common.GeneratedCommentHeader(globalConfig))
 
 	return jenny


internal/jennies/golang/types.go
```diff
@@ -221,3 +259,20 @@ func (formatter *typeFormatter) formatIntersection(def ast.IntersectionType) str
 
 	return buffer.String()
 }
+
+func toOrderedMap(m map[string]interface{}) *orderedmap.Map[string, interface{}] {
```
K-Phoen, line 263: This function isn't specific to Go, so I would move it directly in the `orderedmap` package :)

Maybe as a kind of named constructor?

```go
func FromMap[K constraints.Ordered, V any](original map[K]V) *Map[K, V] {
	orderedMap := New[K, V]()

	keys := make([]K, 0, len(original))
	for key := range original {
		keys = append(keys, key)
	}

	sort.Slice(keys, func(i, j int) bool { return keys[i] < keys[j] })

	for _, k := range keys {
		orderedMap.Set(k, original[k])
	}

	return orderedMap
}
```

internal/jennies/typescript/rawtypes.go
```diff
@@ -351,3 +373,26 @@ func formatValue(val any) string {
 
 	return fmt.Sprintf("%#v", val)
 }
+
+func hasStructDefaults(typeDef ast.Type, defaults any) bool {
+	_, ok := defaults.(map[string]interface{})
+	return ok && typeDef.Kind == ast.KindStruct
+}
+
+func toOrderedMap(value any) *orderedmap.Map[string, interface{}] {
```
K-Phoen, line 382: This function is pretty much the same than the Go one, so I would definitely move it to the `orderedmap` package :D

internal/simplecue/utils.go
```diff
@@ -177,8 +177,30 @@ func cueConcreteToScalar(v cue.Value) (interface{}, error) {
 
 		return values, nil
 	case cue.StructKind:
-		// TODO: this would be used for struct-level default values
-		return nil, nil //nolint: nilnil
+		newMap := make(map[string]interface{})
+		iter, _ := v.Fields(cue.Optional(true), cue.Definitions(true))
```
K-Phoen, line 181: What's the second return value of `v.Fields()`? 
ie: are we silencing an error here? :scream:

internal/simplecue/utils.go
```diff
@@ -177,8 +177,30 @@ func cueConcreteToScalar(v cue.Value) (interface{}, error) {
 
 		return values, nil
 	case cue.StructKind:
-		// TODO: this would be used for struct-level default values
-		return nil, nil //nolint: nilnil
+		newMap := make(map[string]interface{})
```
K-Phoen, line 180: Should we be populating an `orderedmap` here? To guarantee a consistent output of the parser

testdata/jennies/builders/struct_with_defaults/schema.cue
```diff
@@ -0,0 +1,17 @@
...
+#Struct: {
+	allFields: #NestedStruct | *{
+  	stringVal: "hello"
+  	intVal: 3
+  }
+  partialFields: #NestedStruct | *{
+  	intVal: 4
+  }
+  emptyFields: #NestedStruct | *{}
+}
```
K-Phoen, line 17: Should we add a test case with anonymous structs?

```cue
#SomeStruct: {
	complexField: {
		intVal: int64
	} | *{ intVal: 2 }
}
```

internal/jennies/golang/types.go
```diff
@@ -162,6 +164,42 @@ func formatScalar(val any) string {
 	return fmt.Sprintf("%#v", val)
 }
 
+func formatDefaultStruct(refPkg, pkg string, structMap *orderedmap.Map[string, interface{}]) string {
+	starter, format, sep, lastSep, ending := "{\n", "%s: %v", ",\n", ",\n", "}"
+	if pkg != "" {
+		starter, format, sep, lastSep, ending = fmt.Sprintf("New%sBuilder().\n", pkg), "%s(%v)", ".\n", ",\n", ""
```
K-Phoen, line 170: Are we using the package as builder name here? If so, that's probably incorrect.

See https://github.com/grafana/cog/blob/2fd0d68c215e8231b9184d6b2d0d1b05c446fd4c/internal/jennies/golang/templates/builders/builder.tmpl#L18

K-Phoen (commented):
Haven't fetched this branch locally to play with it yet, but here's the first batch of comments :)

internal/simplecue/utils.go
spinillos: I omitted it because we are inside struct clause.

testdata/jennies/builders/struct_with_defaults/schema.cue
spinillos: Added :)

internal/simplecue/utils.go
K-Phoen: And that means we don't have to deal we errors? :thinking: (I'm not super super familiar with how cue works, so it's a real question :D)

spinillos: It returns an error if it isn't a struct:
```
// Fields creates an iterator over v's fields if v is a struct or an error
// otherwise.
```

K-Phoen (approved):

PR #187, opened 2023-12-11
Define veneers for dashboard variables

K-Phoen:
This PR configures a few veneers to improve the API around dashboard variables.

These variables are described by a single model in the schemas, even though different variable types actually use a subset of the fields in that model.

I use veneers to generate a builder per variable type exposed by the schema, and I restrict the options generated for each builder to what Grafana allows us to configure.

### DIFF
--- a/config/dashboard.variables.veneers.common.yaml
+++ b/config/dashboard.variables.veneers.common.yaml
@@ -120,10 +120,10 @@ options:
   - omit: { by_builder: DatasourceVariable.options }
   - rename:
       by_builder: DatasourceVariable.query
-      as: datasourceType
-  # DatasourceType("prometheus") instead of explicit DatasourceType(StringOrAny{String: "prometheus"})
+      as: type
+  # Type("prometheus") instead of explicit Type(StringOrAny{String: "prometheus"})
   - struct_fields_as_arguments:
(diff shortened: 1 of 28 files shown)


config/dashboard.variables.veneers.common.yaml
```diff
@@ -0,0 +1,162 @@
...
+  - omit: { by_builder: DatasourceVariable.datasource }
+  - omit: { by_builder: DatasourceVariable.sort }
+  - omit: { by_builder: DatasourceVariable.refresh }
+  - omit: { by_builder: DatasourceVariable.options }
+  - rename:
+      by_builder: DatasourceVariable.query
+      as: datasourceType
+  # DatasourceType("prometheus") instead of explicit DatasourceType(StringOrAny{String: "prometheus"})
+  - struct_fields_as_arguments:
+      by_builder: DatasourceVariable.datasourceType
```
spinillos, line 126: Should be better `name` or `type`?. Adding datasource its a bit redundant.

K-Phoen: Fair point, I'll change that to `type` :)

spinillos (approved):

PR #195, opened 2023-12-13
Bootstrap support for Python

K-Phoen:
This PR bootstraps types generation in Python.

It's the initial pass, not everything is implemented (even just for types). Most notably:

* no handling of default values
* no constructor is generated for the types
* no support for (de)serialization

My apologies for the big PR, I got carried away once more :|
Exploring it commit by commit should help. Most of the interesting code lives in `internal/jennies/python/rawtypes.go` and `internal/jennies/python/types.go`

But on the bright side, when running this branch on `kind-registry`, [`mypy`](https://mypy.readthedocs.io/) seems to be happy with what we generate :)

### DIFF
--- a/internal/ast/compiler/anonymous_structs_to_named.go
+++ b/internal/ast/compiler/anonymous_structs_to_named.go
@@ -1,0 +1,122 @@
+package compiler
+
+import (
+	"github.com/grafana/cog/internal/ast"
+	"github.com/grafana/cog/internal/tools"
+)
+
+var _ Pass = (*AnonymousStructsToNamed)(nil)
+
+// AnonymousStructsToNamed turns "anonymous structs" into a named object.
+//
+// Example:
+//
+//	```
+//	Panel struct {
+//		Options struct {
+//			Title string
+//		}
+//	}
+//	```
+//
+// Will become:
+//
+//	```
+//	Panel struct {
+//		Options PanelOptions
+//	}
+//
+//	PanelOptions struct {
+//		Title string
+//	}
+//	```
+type AnonymousStructsToNamed struct {
+	newObjects []ast.Object
+}
+
+func (pass *AnonymousStructsToNamed) Process(schemas []*ast.Schema) ([]*ast.Schema, error) {
+	newSchemas := make([]*ast.Schema, 0, len(schemas))
+	for _, schema := range schemas {
+		newSchemas = append(newSchemas, pass.processSchema(schema))
+	}
+
+	return newSchemas, nil
+}
+
+func (pass *AnonymousStructsToNamed) processSchema(schema *ast.Schema) *ast.Schema {
+	pass.newObjects = make([]ast.Object, 0, len(schema.Objects))
+	for _, object := range schema.Objects {
+		newObj := pass.processObject(object)
+		pass.newObjects = append(pass.newObjects, newObj)
+	}
+
+	schema.Objects = pass.newObjects
+
+	return schema
+}
+
+func (pass *AnonymousStructsToNamed) processObject(object ast.Object) ast.Object {
+	newObject := object
+	pkg := object.SelfRef.ReferredPkg
+	parentName := tools.UpperCamelCase(pkg) + tools.UpperCamelCase(object.Name)
+
+	if object.Type.IsAnyOf(ast.KindArray, ast.KindMap, ast.KindDisjunction) {
+		newObject.Type = pass.processType(pkg, parentName, object.Type)
+	}
+
+	if object.Type.IsStruct() {
+		for i, field := range object.Type.AsStruct().Fields {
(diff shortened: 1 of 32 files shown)


internal/ast/compiler/anonymous_structs_to_named.go
```diff
@@ -0,0 +1,125 @@
...
+//	```
+//	Panel struct {
+//		Options PanelOptions
+//	}
+//
+//	PanelOptions struct {
+//		Title string
+//	}
+//	```
+type AnonymousStructsToNamed struct {
```
K-Phoen, line 33: I couldn't find a satisfying way of dealing with "anonymous structs" in Python so I worked around the problem: no more anonymous structs.

internal/ast/compiler/disjunctions.go
```diff
@@ -197,14 +194,6 @@ func (pass *DisjunctionToType) processStruct(schema *ast.Schema, def ast.Type) (
 func (pass *DisjunctionToType) processDisjunction(schema *ast.Schema, def ast.Type) (ast.Type, error) {
 	disjunction := def.AsDisjunction()
 
-	// Ex: type | null
-	if len(disjunction.Branches) == 2 && disjunction.Branches.HasNullType() {
-		finalType := disjunction.Branches.NonNullTypes()[0]
-		finalType.Nullable = true
-
-		return finalType, nil
-	}
```
K-Phoen, line 206: This behavior was interesting to generate cleaner types in Python, so I moved it out into its own compiler pass: [internal/ast/compiler/disjunctions_with_null_to_optional.go](https://github.com/grafana/cog/pull/195/files#diff-30adea340ca2f95cba1a878f783a7eef6d72a2d1aa748c43132785c54f0044fd)

internal/jennies/python/tools.go
```diff
@@ -0,0 +1,85 @@
...
+	if isReservedPythonKeyword(name) {
+		return name + "_val"
+	}
+
+	return name
+}
+
+func isReservedPythonKeyword(input string) bool {
+	// see: https://docs.python.org/3/reference/lexical_analysis.html#keywords
+	return input == "False" ||
```
julienduchesne, line 50: switch case would look better here

testdata/jennies/rawtypes/struct_with_complex_fields.txtar
```diff
@@ -297,3 +299,29 @@ func (resource *StringOrBool) UnmarshalJSON(raw []byte) error {
...
+class StructComplexFieldsSomeStructFieldAnonymousStruct:
+    FieldAny: object
+
+
+class SomeStruct:
+    """
+    This struct does things.
+    """
+
+    FieldRef: 'SomeOtherStruct'
```
julienduchesne, line 316: If you import `from __future__ import annotations`, then you can use the actual ref without quoting

K-Phoen: I tried using that import, but it breaks unions defined as top-level types (with forward references):

```py
# That's broken
ValueMapping = typing.Union[ValueMap | RangeMap | RegexMap | SpecialValueMap]

# That works
ValueMapping = typing.Union['ValueMap' | 'RangeMap' | 'RegexMap' | 'SpecialValueMap']
```

I'm guessing that it's because this import only has an effect on type annotations, and the declaration of `ValueMapping` isn't one.

internal/jennies/python/tools.go
K-Phoen: Done :)

internal/jennies/python/tools.go
```diff
@@ -0,0 +1,60 @@
...
+
+		return fmt.Sprintf("[%s]", strings.Join(items, ", "))
+	}
+
+	return fmt.Sprintf("%#v", val)
+}
+
+func escapeFieldName(name string) string {
+	if isReservedPythonKeyword(name) {
+		return name + "_val"
```
spinillos, line 42: Does it changes the json result? I don't recall if Python uses the field names as json keys.

internal/jennies/python/types.go
```diff
@@ -0,0 +1,284 @@
...
+		return "float"
+	case ast.KindUint8, ast.KindUint16, ast.KindUint32, ast.KindUint64:
+		return "int"
+	case ast.KindInt8, ast.KindInt16, ast.KindInt32, ast.KindInt64:
+		return "int"
+
+	case ast.KindBool:
+		return "bool"
+	default:
+		return string(kind)
```
spinillos, line 252: Maybe use "unknown"? To see that its something that we need to fix.

internal/jennies/python/tools.go
spinillos: I didn't reach this PR yet :)

PR #207, opened 2024-01-09
Java Rawtypes

spinillos:
This is the first iteration for Java types.

Some notes about this:
* Every object in Java is a class, so its a different file ☠️. 
* Generic is set as `Object`.
* Java doesn't allow aliases, so they aren't generated. We use the type of the aliases when we found one.
  * For example: [time.cue](https://github.com/grafana/kind-registry/blob/main/grafana/next/common/time.cue#L10). When `TimeZone` is found, it sets it's type (`string`).
* Constant scalars are generated under `Constants.java` under each package if necessary.
* It doesn't manage anonymous structs yet. It needs a bit more iteration.
* No required/optionals.
  * To use optionals in Java we should use [Optional class](https://docs.oracle.com/javase/8/docs/api/java/util/Optional.html) and I'm not sure if it worth it or not. So at the moment I didn't add it to make it simpler.
* Getters and setters are generated under a flag. They have sense when we don't have any builder to set the values.


internal/jennies/java/rawtypes.go
```diff
@@ -0,0 +1,278 @@
...
+	enum := object.Type.AsEnum()
+	values := make([]EnumValue, len(enum.Values))
+	for i, value := range enum.Values {
+		values[i] = EnumValue{
+			Name:  strings.ToUpper(value.Name),
+			Value: value.Value,
+		}
+	}
+
+	enumType := "Integer"
```
spinillos, line 118: Probably it isn't 100% accurate. 

In [`cue`](https://github.com/grafana/cog/blob/main/internal/simplecue/generator.go#L212) we are setting `ast.KindInt64`. Its a `Long` in Java and it forces to add casting to the values in the generated code. So at the moment using `Integer` makes it simpler and works at the moment.

internal/jennies/java/templates/types/class.tmpl
```diff
@@ -0,0 +1,54 @@
+package {{ .Package }};
+
+{{ .Imports }}
+
+{{- range .Comments }}
+// {{ . }}
+{{- end }}
+public class {{ .Name }}{{ if .Extends }} extends {{ range $i, $e := .Extends }}{{ if gt $i 0 }}, {{ end }}{{ $e }}{{ end }}{{ end }}{{ if .Variant }} implements cog.variants.{{ .Variant }}{{ end }} {
```
spinillos, line 8: Variant sets the package here `cog.variants.` to avoid conflicts with classes with the same name.

internal/jennies/java/jennies.go
```diff
@@ -0,0 +1,59 @@
...
+
+	return newConfig
+}
+
+func New() *Language {
+	return &Language{config: Config{}}
+}
+
+func (language *Language) RegisterCliFlags(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&language.config.GenGettersAndSetters, "getters-and-setters", true, "Generate getters and setters for types")
```
K-Phoen, line 34:
```suggestion
	cmd.Flags().BoolVar(&language.config.GenGettersAndSetters, "java-getters-and-setters", true, "Generate getters and setters for types")
```

Just to make it obvious that this flag is only used for the Java output

internal/jennies/java/rawtypes.go
```diff
@@ -0,0 +1,327 @@
...
+
+type RawTypes struct {
+	config  Config
+	imports *common.DirectImportMap
+
+	typeFormatter *typeFormatter
+}
+
+func (jenny RawTypes) JennyName() string {
+	return "JavaBuilder"
```
K-Phoen, line 22:
```suggestion
	return "JavaRawTypes"
```

internal/jennies/java/rawtypes.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	jenny.typeFormatter = jenny.typeFormatter.withPackageMapper(packageMapper)
+
+	for _, object := range schema.Objects {
+		jenny.imports = NewImportMap()
+		if object.Type.IsMap() {
+			continue
+		}
+		if object.Type.IsScalar() {
```
K-Phoen, line 61: It looks like you're treating these objects are constants: that's not necessarily correct.

This check only tells us that this object aliases a scalar type. For actual constants, we need to do something like `object.Type.IsScalar() && object.Type.AsScalar().IsConcrete()`

internal/jennies/java/rawtypes.go
```diff
@@ -0,0 +1,327 @@
...
+		}
+
+		return jenny.imports.Add(class, pkg)
+	}
+
+	jenny.typeFormatter = jenny.typeFormatter.withPackageMapper(packageMapper)
+
+	for _, object := range schema.Objects {
+		jenny.imports = NewImportMap()
+		if object.Type.IsMap() {
```
K-Phoen, line 58: Does that mean that any object aliasing a "map type" would just be completely omitted? (and if so: is that what we want?)

internal/jennies/java/rawtypes.go
```diff
@@ -0,0 +1,327 @@
...
+		return jenny.formatStruct(pkg, object)
+	case ast.KindEnum:
+		return jenny.formatEnum(pkg, object)
+	case ast.KindRef:
+		return jenny.formatReference(pkg, object)
+	case ast.KindIntersection:
+		return jenny.formatIntersection(pkg, object)
+	}
+
+	return nil, nil
```
K-Phoen, line 104: Shouldn't we return an error if we end up here?
Also: why aren't array and maps included in the switch statement?

internal/jennies/java/rawtypes.go
```diff
@@ -0,0 +1,327 @@
...
+func escapeVarName(varName string) string {
+	if isReservedGoKeyword(varName) {
+		return varName + "Arg"
+	}
+
+	return varName
+}
+
+// nolint: gocyclo
+func isReservedGoKeyword(input string) bool {
```
K-Phoen, line 278:
```suggestion
func isReservedJavaKeyword(input string) bool {
```

internal/jennies/java/rawtypes_test.go
```diff
@@ -0,0 +1,39 @@
...
+		Name: "jennies/JavaRawTypes",
+	}
+
+	jenny := RawTypes{}
+	compilerPasses := New().CompilerPasses()
+
+	test.Run(t, func(tc *txtartest.Test) {
+		req := require.New(tc)
+
+		// We run the compiler passes defined fo Go since without them, we
```
K-Phoen, line 24:
```suggestion
		// We run the compiler passes defined for Java since without them, we
```

internal/jennies/java/rawtypes_test.go
```diff
@@ -0,0 +1,39 @@
...
+		// might not be able to translate some of the IR's semantics into Go.
```
K-Phoen, line 25:
```suggestion
		// might not be able to translate some of the IR's semantics into Java.
```

internal/jennies/java/templates/runtime/variants.tmpl
```diff
@@ -0,0 +1,5 @@
+package cog.variants;
+
+public interface {{ .Variant }} {
+    public void Implements{{ .Variant }}Variant();
```
K-Phoen, line 4: Do we need this empty method in Java? Isn't the fact that a class explicitly has to *implement* the interface enough?

testdata/jennies/rawtypes/disjunctions.txtar
```diff
@@ -360,3 +360,82 @@ func (resource *StringOrBool) UnmarshalJSON(raw []byte) error {
...
+public class RefreshRate extends StringOrBool {
+    
+}
+== disjunctions/SomeStruct.java
+package disjunctions;
+
+
+public class SomeStruct {
+    private String Type;
+    private Object FieldAny;
```
K-Phoen, line 378: If we don't generate getters/setters, shouldn't we set these fields as public? :thinking:

testdata/jennies/rawtypes/enums.txtar
```diff
@@ -213,3 +213,81 @@ const (
...
 
 
+-- out/jennies/JavaRawTypes --
+== enums/Operator.java
+package enums;
+
+
+// This is a very interesting string enum.
+public enum Operator {
+    GREATERTHAN(">"),
```
K-Phoen, line 223: `GREATER_THAN(">")` would look better :innocent:

K-Phoen (approved):
Left a bunch of small comments, but it looks like a very good start!

internal/jennies/java/templates/runtime/variants.tmpl
spinillos: That's true. I didn't recall it.

internal/jennies/java/rawtypes.go
spinillos: Its because their where checked previously and skipped because they were aliases.

internal/jennies/java/rawtypes.go
spinillos: I'd say that I did it like this to simplify it.

spinillos: I just added arrays for the same reason.

PR #210, opened 2024-01-11
Refactor defaults/empty values management in TS

K-Phoen:
Small refactor for #175


K-Phoen:
`CI / Run examples` is failing because this PR (and it's parent) are quite out-of-date with main. Rebasing both would fix this failure.

PR #213, opened 2024-01-12
Fix map values formatting for initialization safeguards in builders

K-Phoen:
I noticed that some initialization safeguards in typescript could be generated incorrectly:

```typescript
this.internal.options.legend = map[string]interface {}{"show":true};
```

This PR addresses that.

ToDo:

* [x] Add a txtar test case

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -51,6 +51,51 @@ jobs:
       - name: Tests
         run: make tests
 
+  registry:
+    name: Generate registry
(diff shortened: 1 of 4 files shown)


K-Phoen:
I also added a CI job to run cog on the kind-registry and lint the results.

Turns out it fails because of a bug in Typescript \o/

K-Phoen:
The new CI job will be fixed by https://github.com/grafana/cog/pull/215

spinillos (approved):

PR #218, opened 2024-01-18
Python/json unmarshal

K-Phoen:
This PR kickstarts the implementation of unmarshaling JSON data into Python models.

The static `from_json()` method introduced here isn't finished yet:
* panels and dataqueries aren't handled properly
* disjunctions aren't handled properly
* probably other things that I haven't noticed yet 🫠

To keep the diff size reasonable, these will be implemented in subsequent PRs

### DIFF
--- a/config/dashboard.veneers.python.yaml
+++ b/config/dashboard.veneers.python.yaml
@@ -19,3 +19,11 @@ options:
   # Time(from, to) instead of time(struct {From string `json:"from"`, To string `json:"to"`}{From: "lala", To: "lala})
   - struct_fields_as_arguments:
       by_name: Dashboard.time
+
+  ##############
+  #   Panels   #
+  ##############
+
+  # WithOverride(matcher, properties) instead of WithOverride(struct{...})
+  - struct_fields_as_arguments:
+      by_name: Panel.withOverride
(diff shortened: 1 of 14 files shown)


testdata/jennies/rawtypes/struct_with_optional_fields.txtar
```diff
@@ -254,6 +261,24 @@ class SomeStruct:
             payload["FieldAnonymousStruct"] = self.field_anonymous_struct.to_json()
         return payload
 
+    @classmethod
+    def from_json(cls, data: dict[str, typing.Any]) -> typing.Self:
+        args = {
```
julienduchesne, line 266: Should we format with a tool like "black"? That would remove weird new spaces like this and probably make the generation slightly easier (less important to have correct spacing)

julienduchesne (approved):

testdata/jennies/rawtypes/struct_with_optional_fields.txtar
K-Phoen: As an optional post-processing step, maybe?

My initial gut feeling is that requiring third-party tools to be installed to run cog is going to be painful (especially as we add more languages)

PR #219, opened 2024-01-19
Split txtar fixtures on disk

K-Phoen:
Txtar tests are nice but each new language cog supports makes them harder to read.

This PR implements a proposal made by @spinillos: let's give up on having a test case represented by a single txtar file, and instead fully write our *golden files* on disk.

The vast majority of changes made here are related to txtar files being split into several files. The only interesting code change is the introduction of `internal/testutils/jennytestsuite.go` instead of `internal/txtartest/txtar.go`


julienduchesne (approved):
Same number of tests as before 👍. I like this. I'm not a big fan of composite formats that aren't easily formattable

PR #223, opened 2024-01-22
Split the disjunction discriminator and mapping inference code in a standalone compiler pass

K-Phoen:
This PR splits the `DisjunctionToType` compiler pass even further, isolating the code responsible for inferring the discriminator field and mapping to use in its own pass.

I am doing this for two reasons:
* to make these passes less hard to read and understand
* to enable unmarshalling from JSON easier in languages like Python that support disjunctions but still need to know the discriminator & mapping to unmarshal correctly

Note: this PR doesn't change the code generated by cog

### DIFF
--- a/internal/ast/compiler/disjunctions_infer_mapping.go
+++ b/internal/ast/compiler/disjunctions_infer_mapping.go
@@ -1,0 +1,261 @@
+package compiler
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/grafana/cog/internal/ast"
+)
+
+var _ Pass = (*DisjunctionInferMapping)(nil)
+
+var ErrCanNotInferDiscriminator = errors.New("can not infer discriminator mapping")
+
+// DisjunctionInferMapping infers the discriminator field and mapping used to
+// describe a disjunction of references.
+// See https://swagger.io/docs/specification/data-models/inheritance-and-polymorphism/
+type DisjunctionInferMapping struct {
+}
+
+func (pass *DisjunctionInferMapping) Process(schemas []*ast.Schema) ([]*ast.Schema, error) {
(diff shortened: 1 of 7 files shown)


internal/ast/compiler/disjunctions_infer_mapping.go
```diff
@@ -0,0 +1,261 @@
...
+}
+
+func (pass *DisjunctionInferMapping) ensureDiscriminator(schema *ast.Schema, def *ast.DisjunctionType) (*ast.DisjunctionType, error) {
+	// discriminator-related data was set during parsing: nothing to do.
+	if def.Discriminator != "" && len(def.DiscriminatorMapping) != 0 {
+		return def, nil
+	}
+
+	if def.Discriminator == "" {
+		def.Discriminator = pass.inferDiscriminatorField(schema, def)
```
spinillos, line 142: Shouldn't we return an error if we don't find the discriminator?

internal/ast/compiler/disjunctions_infer_mapping.go
```diff
@@ -0,0 +1,261 @@
...
+
+	// Identify candidates from each branch
+	for _, branch := range def.Branches {
+		referredType, found := schema.Resolve(branch)
+		if !found {
+			continue
+		}
+
+		typeName := branch.AsRef().ReferredType
+		structType := referredType.AsStruct()
```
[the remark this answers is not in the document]
K-Phoen: Yeah, that's correct. I added a check to guard against that :)

internal/ast/compiler/disjunctions_infer_mapping.go
K-Phoen: All the error handling is done in the `buildDiscriminatorMapping()` function called a few lines below

Issue #228, opened 2024-01-25
Add type field to Argument constant


spinillos:
This issue is to investigate if there is a easy way to do this change:

Currently we have:
```go
type AssignmentValue struct {
	Argument *Argument           `json:",omitempty"`
	Constant any                 `json:",omitempty"`
	Envelope *AssignmentEnvelope `json:",omitempty"`
}
```
And we want to update with something like:
```go
type AssignmentValue struct {
	Argument *Argument           `json:",omitempty"`
	Constant *Constant                 `json:",omitempty"`
	Envelope *AssignmentEnvelope `json:",omitempty"`
}

type Constant {
   Value any
   Type Type
}
```

Why? Argument Constant doesn't have any information about the type of the field that we are we setting.

From most of the languages ins't a big deal, but for Java enums are classes and we couldn't set a String or an Integer directly here.

K-Phoen:
Isn't the information found in [`Assignment.Path`](https://github.com/grafana/cog/blob/c1c398324462034cff738b92db82f546861f3bb5/internal/ast/builder.go#L295) enough?

PR #229, opened 2024-01-25
Java: Generate builders

spinillos:
It generates the builders for Java. Could be done better? Of course.
Some parts are a bit odd since I needed to deal with path positions to know if I should set a direct assignation, a getter or a setter.

To improve enums this issue was added: https://github.com/grafana/cog/issues/228. At least to investigate it if we could do it a bit more elegant.

Note: This PR doesn't include error handler, defaults or tests.


julienduchesne:
Can you rebase? This doesn't run for me locally

julienduchesne (commented):
Needs a rebase. Submitting a review so it gets off my queue until it's rebased 😄. I'd be happy to try it out once that's done

spinillos:
@julienduchesne I just update the branch. I was completely focus in other stuff and I forgot that 😓

PR #233, opened 2024-01-31
Orderedmap utils

K-Phoen:
This PR adds a few functions to make working with our ordered map implementation a bit easier.

They're not used anywhere yet, but I'll need these in PRs that will be coming soon :) (in which I'll slightly refactor the way we store objects and struct fields in our IR)


Duologic (commented):
Have you considered using an existing package for this? Quick search surfaced this: https://pkg.go.dev/github.com/wk8/go-ordered-map

PR #234, opened 2024-02-01
Support packages with dashes in Go

K-Phoen:
Our jennies do not sanitize package names enough, so if a package contains dashes it will be rendered as-is in the generated code, which is a problem for most languages.

Starting with Go, let's fix that!

### DIFF
--- a/testdata/jennies/builders/package-with-dashes/PythonBuilder/builders/builder-pkg.py
+++ b/testdata/jennies/builders/package-with-dashes/PythonBuilder/builders/builder-pkg.py
@@ -1,0 +1,19 @@
+import typing
+from ..cog import builder as cogbuilder
+from ..models import with-dashes
+
+
+class SomeNiceBuilder(cogbuilder.Builder[with-dashes.SomeStruct]):    
+    __internal: with-dashes.SomeStruct
+
+    def __init__(self):
+        self.__internal = with-dashes.SomeStruct()
+
+    def build(self) -> with-dashes.SomeStruct:
+        return self.__internal    
+    
+    def title(self, title: str) -> typing.Self:        
+        self.__internal.title = title
+    
+        return self
+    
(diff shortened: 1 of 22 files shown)


K-Phoen:
Relates to #235

testdata/jennies/builders/package-with-dashes/TypescriptBuilder/src/builder-pkg/somenicebuilder_builder_gen.ts
```diff
@@ -0,0 +1,19 @@
+import * as cog from '../cog';
+import * as with-dashes from '../with-dashes';
```
julienduchesne, line 2: I think this should be replaced too. I think the same logic as go would look better than a dash here. Or even better would be to capitalize
```suggestion
import * as withDashes from '../with-dashes';
```

testdata/jennies/builders/package-with-dashes/PythonBuilder/builders/builder-pkg.py
```diff
@@ -0,0 +1,19 @@
+import typing
+from ..cog import builder as cogbuilder
+from ..models import with-dashes
+
+
+class SomeNiceBuilder(cogbuilder.Builder[with-dashes.SomeStruct]):
```
julienduchesne, line 6: The pythonic equivalent would be an underscore. I think what would work best for python is the same logic as golang but instead of replacing by an empty space, replace with an underscore

```suggestion
class SomeNiceBuilder(cogbuilder.Builder[with_dashes.SomeStruct]):    
```

julienduchesne (approved):
Approved because I assume you will do other languages in future PRs! Empty space seems like the best option to replace unwanted chars, indeed

K-Phoen:
Yup, similar fixes are needed for the other languages too. I'll keep track of the remaining work in https://github.com/grafana/cog/issues/235 :)

PR #240, opened 2024-02-07
Add missing "interval" field in prometheus dataquery

K-Phoen:
Relates to https://github.com/grafana/grafana-foundation-sdk/issues/36

The CUE schema doesn't contain this field, so until it's updated we can *artificially* define it with a compiler pass.

Cf grafana/grafana for its definition: https://github.com/grafana/grafana/blob/9c0501a1674ba66169e082984f36534427a0eaf8/public/app/plugins/datasource/prometheus/types.ts#L16

### DIFF
--- a/internal/ast/compiler/prometheusdataquery.go
+++ b/internal/ast/compiler/prometheusdataquery.go
@@ -1,0 +1,45 @@
+package compiler
+
+import (
+	"github.com/grafana/cog/internal/ast"
+)
+
+var _ Pass = (*PrometheusDataquery)(nil)
+
+// PrometheusDataquery rewrites the definition of the prometheus.Dataquery type and adds a few missing fields.
+// Note: this pass is meant to be removed once the schema is up-to-date.
+type PrometheusDataquery struct {
+}
+
+func (pass *PrometheusDataquery) Process(schemas []*ast.Schema) ([]*ast.Schema, error) {
+	for i, schema := range schemas {
+		schemas[i] = pass.processSchema(schema)
+	}
+
+	return schemas, nil
+}
+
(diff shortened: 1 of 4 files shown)


internal/ast/compiler/prometheusdataquery.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"github.com/grafana/cog/internal/ast"
+)
+
+var _ Pass = (*PrometheusDataquery)(nil)
+
+// PrometheusDataquery rewrites the definition of the prometheus.Dataquery type and adds a few missing fields.
+// Note: this pass is meant to be removed once the schema is up-to-date.
+type PrometheusDataquery struct {
+}
```
julienduchesne, line 12: Extreme nit
```suggestion
type PrometheusDataquery struct {}
```

internal/ast/compiler/prometheusdataquery.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+
+	return schemas, nil
+}
+
+func (pass *PrometheusDataquery) processSchema(schema *ast.Schema) *ast.Schema {
+	for i, object := range schema.Objects {
+		if schema.Package == prometheusPackage && object.Name == prometheusDataqueryObject {
+			schema.Objects[i] = pass.processDataquery(object)
+			continue
```
julienduchesne, line 26: That continue doesn't do anything, right?

julienduchesne (approved):

internal/ast/compiler/prometheusdataquery.go
K-Phoen: Yeah, just a habit :)

PR #242, opened 2024-02-07
Fix go disjunctions marshalling

K-Phoen:
Fixes https://github.com/grafana/grafana-foundation-sdk/issues/38

This PR fixes two bugs related to how disjunctions are handled and marshaled to JSON in go.

The first commit ensures that a struct resulting from a disjunction is marshaled correctly, irrelevantly of whether you use a pointer receiver or not.

The second commit ensures that type aliases are actually defined as aliases (as opposed to a new type). This is to ensure that the correct `MarshalJSON()` function will be called when marshaling.

### DIFF
--- a/testdata/jennies/rawtypes/refs/GoRawTypes/refs/types_gen.go
+++ b/testdata/jennies/rawtypes/refs/GoRawTypes/refs/types_gen.go
@@ -8,7 +8,7 @@ type SomeStruct struct {
 	FieldAny any `json:"FieldAny"`
 }
 
-type RefToSomeStruct SomeStruct
+type RefToSomeStruct = SomeStruct
 
-type RefToSomeStructFromOtherPackage otherpkg.SomeDistantStruct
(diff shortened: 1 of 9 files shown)


testdata/jennies/rawtypes/refs/GoRawTypes/refs/types_gen.go
```diff
@@ -8,7 +8,7 @@ type SomeStruct struct {
 	FieldAny any `json:"FieldAny"`
 }
 
-type RefToSomeStruct SomeStruct
+type RefToSomeStruct = SomeStruct
 
-type RefToSomeStructFromOtherPackage otherpkg.SomeDistantStruct
+type RefToSomeStructFromOtherPackage = otherpkg.SomeDistantStruct
```
julienduchesne, line 13: huh, didn't know you could define types like that

julienduchesne (approved):

PR #264, opened 2024-02-22
Better support for empty objects/any types in jsonschema

K-Phoen:
Fixes a few edge cases where our jsonschema parser incorrectly interpreted object definitions.

### DIFF
--- a/testdata/jsonschema/basic_object/GenerateAST/ir.json
+++ b/testdata/jsonschema/basic_object/GenerateAST/ir.json
@@ -137,6 +137,63 @@
               "Required": false
             },
             {
+              "Name": "objectAdditionalPropertiesAny",
+              "Type": {
+                "Kind": "scalar",
+                "Nullable": false,
+                "Scalar": {
+                  "ScalarKind": "any"
+                }
+              },
+              "Required": false
+            },
+            {
+              "Name": "objectAdditionalPropertiesTyped",
+              "Type": {
+                "Kind": "map",
+                "Nullable": false,
+                "Map": {
+                  "IndexType": {
+                    "Kind": "scalar",
+                    "Nullable": false,
+                    "Scalar": {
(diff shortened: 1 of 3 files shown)


testdata/jsonschema/basic_object/GenerateAST/ir.json
```diff
@@ -136,6 +136,63 @@
...
+                    "Scalar": {
+                      "ScalarKind": "string"
+                    }
+                  }
+                }
+              },
+              "Required": false
+            },
+            {
+              "Name": "objectEmptyPropertiesIsAny",
```
julienduchesne, line 175: That seems to be the same as `objectNoPropertiesIsAny` but it should in fact be an empty object, with no properties right. This seems to allow everything now

testdata/jsonschema/basic_object/GenerateAST/ir.json
```diff
@@ -136,6 +136,63 @@
               },
               "Required": false
             },
+            {
+              "Name": "objectAdditionalPropertiesAny",
+              "Type": {
+                "Kind": "scalar",
+                "Nullable": false,
+                "Scalar": {
+                  "ScalarKind": "any"
```
julienduchesne, line 145: This doesn't seem like it's doing the expected thing right? It should be a map of string -> object, right?

Example:
```
{
	"objectAdditionalPropertiesAny": {
    "hello": "hello"
    }
}
```
fails with
```
Invalid type. Expected Object but got String. Path 'objectAdditionalPropertiesAny.hello', line 3, position 20.
```
when I try to validate it. The AST doesn't seem to reflect that behavior

K-Phoen: You're definitely correct! :facepalm: 

I'll close this PR :+1:

PR #274, opened 2024-02-26
Fix go custom marshal function to correctly handle null cases

K-Phoen:
Custom marshal functions need to return valid JSON.

### DIFF
--- a/testdata/jennies/rawtypes/disjunctions/GoJSONMarshalling/disjunctions/types_json_marshalling_gen.go
+++ b/testdata/jennies/rawtypes/disjunctions/GoJSONMarshalling/disjunctions/types_json_marshalling_gen.go
@@ -8,7 +8,7 @@ func (resource StringOrBool) MarshalJSON() ([]byte, error) {
 		return json.Marshal(resource.Bool)
 	}
 
-	return nil, nil
+	return []byte("null"), nil
 }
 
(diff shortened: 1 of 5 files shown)


testdata/jennies/rawtypes/disjunctions/GoJSONMarshalling/disjunctions/types_json_marshalling_gen.go
```diff
@@ -8,7 +8,7 @@ func (resource StringOrBool) MarshalJSON() ([]byte, error) {
 		return json.Marshal(resource.Bool)
 	}
 
-	return nil, nil
+	return []byte("null"), nil
```
julienduchesne, line 11: I feel like this should be an error instead of returning null

K-Phoen: Valid point. Returning null in these cases makes the generated code very lenient: it will marshal to "null" some fields that the schema indicated as required but were not provided with a value (default or otherwise).

I updated this PR to return an error instead, and I'll open another PR to address https://github.com/grafana/grafana-foundation-sdk/issues/50 specifically :)

K-Phoen:
https://github.com/grafana/cog/pull/288 should fix the failing "Run examples" CI step

spinillos (approved):

Issue #235, opened 2024-02-01
Package names with dashes result in invalid code


K-Phoen:
Our jennies do not sanitize package names enough, so if a package contains dashes it will be rendered as-is in the generated code, which is a problem for most languages.

* [ ] Python
* [ ] Java
* [ ] Typescript
* [ ] Go

PR #276, opened 2024-02-26
Correctly format package and object names with dashes in TS

K-Phoen:
Part of #235

This PR updates our Typescript jenny to properly escape package and object names.

Note: filenames for types and builder change, but this will have no impact on end-users since we expose things via `index.ts` files.


julienduchesne (approved):
👍 Looks better with proper casing!

PR #285, opened 2024-03-11
Escape enum members names "nan" in TS as it is an invalid identifier

K-Phoen:
`ts-node` in the CI pipeline picked that up in my previous PR. Not sure why this didn't pop up before :shrug:


K-Phoen:
The `CI / Run examples` step is failing because the prometheus query schema has been removed from grafana and is no longer published in the `next` folder of the kind-registry :|

I'll figure out something to work around that in another PR

PR #290, opened 2024-03-15
Bootstrap jsonschema jenny

K-Phoen:
Being able to generate JSONSchema/OpenAPI from a mixed set of input schemas could be useful, especially to remove grafonnet's dependency on grok.

ToDo:

* [x] identify schema entrypoint
* [x] properly handle comments + compiler passes trail
* [x] resolve + expand foreign references
* [x] ~~check if this could be useful to replace grok-generated schemas in grafonnet (if so: bye-bye grok 🫡)~~ @Duologic agrees: this could work.
* [x] constraints
* [x] tests!
  * [x] ensure that generated schemas are valid
  * [x] ensure that cog can parse the jsonschemas it generates
  * [x] txtar tests

### DIFF
--- a/internal/jennies/all.go
+++ b/internal/jennies/all.go
@@ -9,6 +9,7 @@ import (
 	"github.com/grafana/cog/internal/jennies/golang"
 	"github.com/grafana/cog/internal/jennies/java"
 	"github.com/grafana/cog/internal/jennies/jsonschema"
+	"github.com/grafana/cog/internal/jennies/openapi"
 	"github.com/grafana/cog/internal/jennies/python"
 	"github.com/grafana/cog/internal/jennies/typescript"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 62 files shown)


internal/jennies/all.go
```diff
@@ -50,8 +52,10 @@ func (languageJennies LanguageJennies) AsLanguageRefs() []string {
 func All() LanguageJennies {
 	return LanguageJennies{
 		golang.LanguageRef:     golang.New(),
+		java.LanguageRef:       java.New(),
```
malcolmholmes, line 55: are we adding Java now?

malcolmholmes (approved):

internal/jennies/all.go
K-Phoen: It was already in the list, the items just got reordered a bit :)

Btw: this list is the list of supported languages, it doesn't mean Java will be published to the foundation sdk straight away

PR #297, opened 2024-03-20
Python/packaging

K-Phoen:
This PR kickstarts package generation for Python.

Running `python3 -m build` in the `python` folder of grafana-foundation-sdk produces a package that I could install and use in a venv.

How to build the package the package:

* `cd grafana-foundation-sdk/python`
* `python3 -m pip install --upgrade build`
* `python3 -m build`
* `ls -lh ./dist`

Note: having never built a python package before, I don't really know what I'm doing and just followed https://packaging.python.org/en/latest/tutorials/packaging-projects/ :sweat_smile: 

ToDo:

* [x] add quick example in the package's README
* [x] use the correct version in the package's definition + README

### DIFF
--- a/repository_templates/python/.github/workflows/python-release.yaml
+++ b/repository_templates/python/.github/workflows/python-release.yaml
@@ -6,7 +6,7 @@ on:
       - 'python/grafana_foundation_sdk/**'
 
 env:
-  PYTHON_VERSION: '3.11'
+  PYTHON_VERSION: '3.12'
 
 jobs:
   release:
--- a/repository_templates/python/.github/workflows/python-release.yaml
+++ b/repository_templates/python/.github/workflows/python-release.yaml
@@ -1,0 +1,48 @@
+name: Python Release
+on:
+  pull_request_target:
+    types: [ closed ]
+    paths:
+      - 'python/grafana_foundation_sdk/**'
+
+env:
+  PYTHON_VERSION: '3.11'
+
+jobs:
+  release:
+    if: github.event.pull_request.merged == true && github.base_ref == '{{ .Extra.ReleaseBranch }}'
+
+    name: Build and release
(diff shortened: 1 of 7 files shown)


julienduchesne (approved):
LGTM!

repository_templates/python/.github/workflows/python-release.yaml
```diff
@@ -0,0 +1,48 @@
+name: Python Release
+on:
+  pull_request_target:
+    types: [ closed ]
+    paths:
+      - 'python/grafana_foundation_sdk/**'
+
+env:
+  PYTHON_VERSION: '3.11'
```
julienduchesne, line 9: 3.12?

repository_templates/python/.github/workflows/python-release.yaml
```diff
@@ -0,0 +1,48 @@
...
+        shell: bash
+        working-directory: ./python
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Setup Python {{ `${{ env.PYTHON_VERSION }}` }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: {{ `${{ env.PYTHON_VERSION }}` }}
```
julienduchesne, line 37:
```suggestion
          python-version: "{{ env.PYTHON_VERSION }}"
```
This should work 🤔

K-Phoen: Nope: this file is a Go template, that's why I'm escaping this bit :) (it would otherwise fail since `env.PYTHON_VERSION` doesn't exist in the template's context)

repository_templates/python/.github/workflows/python-release.yaml
K-Phoen: Updated :+1:

PR #299, opened 2024-03-22
Fix/gridpos calculations

K-Phoen:
This PR fixes calculations made to position panels on the grid.
It also implements these calculations to panels defined within a row, in addition to the ones added directly to the dashboard.

Should fix https://github.com/grafana/grafana-foundation-sdk/issues/56


julienduchesne (approved):
LGTM! I wish there was a way to represent this in a language agnostic way

PR #304, opened 2024-04-03
Cleanup imports management in Python

K-Phoen:
Mostly some renamings, to make the imports management more understandable.

Imports are also now sorted, to ensure a reproducible output

### DIFF
--- a/internal/jennies/common/importmap.go
+++ b/internal/jennies/common/importmap.go
@@ -12,6 +12,7 @@ type ImportMapConfig[M any] struct {
 	Formatter           func(importMap M) string
 	AliasSanitizer      func(string) string
 	ImportPathSanitizer func(string) string
+	CurrentPkg          string
 }
 
 type ImportMapOption[M any] func(importMap *ImportMapConfig[M])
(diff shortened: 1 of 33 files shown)


julienduchesne:
It is indeed sorted by the import name, but that's not the standard python way to sort imports

Here's an example of how it should be sorted (as `isort` sorts them):
```
# future imports
from __future__ import annotations

# standard library
import re
import sys
from datetime import date, datetime, timedelta
from pathlib import Path
from unittest import expectedFailure, TestCase, skip

# third-party
import requests
from attr import dataclasses
from honesty.api import download_many

# first-party
from something import other_function, some_function
from . import some_module
from .other_module import SomeClass, some_thing, TestFixture
```

It's sorted: type of import -> import lib -> imported symbol

julienduchesne (commented):
Commenting to remove the review request (posted my previous comment in the wrong place) 😄. Re-request review when ready!

K-Phoen:
Old and unfinished PR. Closing this, and let's see if I feel the need to cleanup python imports again :sweat_smile:

PR #305, opened 2024-04-03
Use "RenameNumericEnumValues" compiler pass for python

K-Phoen:
Needed for schemas that don't provide names for enum members.


julienduchesne:
Doesn't make a difference in the generated diff. Expected? 👀

K-Phoen:
That's because enum members in CUE have names :)

If we were generating code from jsonschema/OpenAPI, it would make a difference

julienduchesne (approved):

PR #312, opened 2024-04-09
Skip foundation-sdk preview when is dependabot

spinillos:
It skips foundation-sdk preview for `dependabot` PR's since they fail because a lack of authorization.


K-Phoen (approved):
:rocket:

PR #315, opened 2024-04-09
Attempt at fixing dependabot PRs

julienduchesne:
PRs like this: https://github.com/grafana/cog/pull/310 fail because of missing permissions. We want the workflow to run but just not post the comments


spinillos (approved):
😅

PR #327, opened 2024-04-19
Upgrade `kin-openapi` library

julienduchesne:
Supersedes https://github.com/grafana/cog/pull/309 
The type is now an array, so we should handle that

### DIFF
--- a/internal/openapi/utils.go
+++ b/internal/openapi/utils.go
@@ -37,6 +37,7 @@ func getEnumType(t string) (ast.Type, error) {
 }
 
 func getConstraints(schema *openapi3.Schema) []ast.TypeConstraint {
+	schemaType := getSchemaType(schema)
 	constraints := make([]ast.TypeConstraint, 0)
 
 	if schema.MinLength > 0 {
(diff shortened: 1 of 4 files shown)


internal/openapi/utils.go
```diff
@@ -95,3 +96,10 @@ func getArgs(v *float64, t string) []any {
 func isRef(ref string) bool {
 	return ref != "" && strings.ContainsAny(ref, "#")
 }
+
+func getSchemaType(schema *openapi3.Schema) string {
+	if schema.Type != nil && len(*schema.Type) == 1 {
+		return (*schema.Type)[0] // Handle multiple types?
```
K-Phoen, line 102: Maybe we should do something like https://github.com/grafana/cog/blob/90dc057b38f3028d041b63a0e5ebc6c9d8e15ec5/internal/jsonschema/generator.go#L148-L149?

(not in the scope of this PR though, we should just create an issue for it)

K-Phoen (commented):
Did you commit the vendors by mistake? :thinking:

K-Phoen (approved):

Issue #328, opened 2024-04-27
Unexpected end of JSON errors for optional DataqueryArray types in generated Go code


andybradshaw:
### Issue Summary

Generated go code for optional Dataquery (and DataqueryArray) types doesn't perform any checks before attempting to deserialize, which can lead to `panic: unexpected end of JSON input` errors.

[Cue type definition](https://github.com/grafana/kind-registry/blob/6c763e7d2869a024ce499e3b78ec427eacc98824/grafana/next/core/dashboard/dashboard.cue#L540-L541):
```
// Depends on the panel plugin. See the plugin documentation for details.
targets?: [...#Target]
```

[Generated deser code](https://github.com/grafana/grafana-foundation-sdk/blob/8fd713f5bd12d1b1ff16b929c44266dbd1aa699d/go/dashboard/types_json_marshalling_gen.go#L195):
```golang
dataqueryTypeHint := ""
if resource.Datasource != nil && resource.Datasource.Type != nil {
	dataqueryTypeHint = *resource.Datasource.Type
}

targets, err := cog.UnmarshalDataqueryArray(fields["targets"], dataqueryTypeHint)
if err != nil {
	return err
}
resource.Targets = targets
```
### Minimal Issue Reproduction

I've included a minimal(-ish) repro based on some dashboards that we have in production. While trying to create a repro locally, it seems that this is not the default panel json that gets created on the latest text panel version (it includes the `datasource` and `targets` fields), but seems like something that the generated code can be defensive against. 

Dashboard json:
```json
{
  "annotations": {
    "list": [
      {
        "builtIn": 1,
        "datasource": {
          "type": "grafana",
          "uid": "-- Grafana --"
        },
        "enable": true,
        "hide": true,
        "iconColor": "rgba(0, 211, 255, 1)",
        "name": "Annotations & Alerts",
        "type": "dashboard"
      }
    ]
  },
  "editable": true,
  "fiscalYearStartMonth": 0,
  "graphTooltip": 0,
  "id": 1,
  "links": [],
  "panels": [
    {
      "gridPos": {
        "h": 8,
        "w": 12,
        "x": 0,
        "y": 0
      },
      "id": 1,
      "options": {
        "code": {
          "language": "plaintext",
          "showLineNumbers": false,
          "showMiniMap": false
        },
        "content": "test content",
        "mode": "markdown"
      },
      "pluginVersion": "10.4.2",
      "type": "text"
    }
  ],
  "schemaVersion": 39,
[...]

PR #329, opened 2024-04-27
Update json marshalling of dataquery types

andybradshaw:
Possible fix for https://github.com/grafana/cog/issues/328, this change adds a presence check for the field before delegating to the `UnmarshalDataquery(Array)` function. An example of this applied to the kind-registry can be found [here](https://github.com/andybradshaw/grafana-foundation-sdk/blob/a49c19b26112212583151d32828235f633ed148e/go/dashboard/types_json_marshalling_gen.go#L195-L201):

```golang
if fields["targets"] != nil {
	targets, err := cog.UnmarshalDataqueryArray(fields["targets"], dataqueryTypeHint)
	if err != nil {
		return err
	}
	resource.Targets = targets
}
```

### DIFF
--- a/internal/jennies/golang/jsonmarshalling.go
+++ b/internal/jennies/golang/jsonmarshalling.go
@@ -315,21 +315,25 @@ dataqueryTypeHint = *resource.%[1]s.Type
 	if field.Type.IsArray() {
 		return fmt.Sprintf(`
 	%[3]s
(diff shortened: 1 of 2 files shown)


K-Phoen (approved):
LGTM :ok_hand: 

Thanks for the PR!

Issue #360, opened 2024-05-08
Create Java code generator


spinillos:
This issue is to track Java builders. The common pattern in Java is to have the builder inside the class, so you can access to the builder doing `MyClass.Builder().blah(...).build()`. An example for this pattern is [this](https://www.digitalocean.com/community/tutorials/builder-design-pattern-in-java#builder-design-pattern-in-java).

But we are going to do it with some constraints to make it easier:
* All fields and functions are going to be public.
  * Java "way" is to use getters/setters, but it does the generator a bit more complex.
  * We can add getters/setters at any point if people starts to ask for that.
* Panels don't have a specific builder, they have a "generic" one adding their extra information from `Options` and `FieldConfig`, so they aren't going to be inside other class.

Tasks:

* [x] Create (base) types for Java. https://github.com/grafana/cog/pull/207
* [x] Create (base) builders for Java. https://github.com/grafana/cog/pull/355
* [x] Create builders for Panels. https://github.com/grafana/cog/pull/364
* [x] Add defaults in builders. https://github.com/grafana/cog/pull/363
* [x] Make Dataquerys and Panels extensible.
* [x] Error management. https://github.com/grafana/cog/pull/376
* [x] Marshall models properly (We could have names modified because they are reserved keywords)
* [x] Create a veneer to delete extensions when the class doesn't add extra fields. https://github.com/grafana/cog/pull/429
* [ ] Create Java linter in CI
* [x] Register into Maven to use the library as a dependency for Maven and Gradle.
  * Useful information for [publishing](https://github.com/grafana/grafana-opentelemetry-starter/blob/main/build.gradle#L76-L148)
* [x] Tests
  * Panel Buidlers
  * Array/Map references

PR #363, opened 2024-05-08
Java: Add default values for builders

spinillos:

It adds defaults in Builders. Its assuming like other languages that the default list are string since we don't have more cases at the moment.

It fixes some issues added in the previous PR: https://github.com/grafana/cog/pull/355.

_Edit_: The tests aren't applying the veneers and anonymous structs are generated as a generic (Object). To fix it in the tests when we detect a struct as argument, it creates an empty object. Veneers in Java names the anonymous structs to remove complexity.

_Edit_: Defaults are applied using the function instead of set directly the value to avoid problems properties or generics.


julienduchesne:
Can you post a before-after diff of the generated java files? I have trouble reviewing these PRs without examples

spinillos:
@julienduchesne Some examples about the defaults are: [this](https://github.com/grafana/cog/pull/363/files#diff-350bf9a458dadf5b30ec530660e013af0f9c33d6cf71ca55f65ccab6ac6e10fbR16-R19) and [this](https://github.com/grafana/cog/pull/363/files#diff-d24b024fa03915ad74ad425eb05b6c6d00d03bf1306b2ee7b0da71092a2ecf75R16-R19). The rest of changes there are the fixes.

Added a note in the description.

K-Phoen (approved):

PR #364, opened 2024-05-10
Java: Generate builders for panels

spinillos:
Contributes to https://github.com/grafana/cog/issues/360

It generates panel's builders for Panels. 

It adds a bit "hacky" things to make it works assuming that if there is a builder called "Panel" under a package, it needs a builder. Dashboard one is skipped since it already has a "Panel" builder and we could have file conflicts if we generate it twice.
 
Most of the changes in the template is to able to do cast to generic objects to be able to set the values. Also the functions that do the stuff have comments to be able to understand what they are doing if we need to re-check them in the future.


spinillos:
The error is fixed in the other PR.

PR #365, opened 2024-05-10
Java: Fix empty enum values and escape reserved words

spinillos:

* It sets default "None" value when enum name doesn't exist. It generates the model like in Go.
* It escapes reserved words from paths to make it works.

### DIFF
--- a/internal/jennies/java/rawtypes.go
+++ b/internal/jennies/java/rawtypes.go
@@ -118,15 +118,15 @@ func (jenny RawTypes) formatEnum(pkg string, object ast.Object) ([]byte, error) 
 	var buffer strings.Builder
 
 	enum := object.Type.AsEnum()
-	values := make([]EnumValue, 0, len(enum.Values))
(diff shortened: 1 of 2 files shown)


internal/jennies/java/rawtypes.go
```diff
@@ -120,6 +120,9 @@ func (jenny RawTypes) formatEnum(pkg string, object ast.Object) ([]byte, error)
 	enum := object.Type.AsEnum()
 	values := make([]EnumValue, len(enum.Values))
 	for i, value := range enum.Values {
+		if value.Name == "" {
+			value.Name = "None"
```
malcolmholmes, line 124: Isn't "None" a Python concept?

malcolmholmes (approved):

internal/jennies/java/rawtypes.go
spinillos: Maybe... but the name that we are using in Golang is "None" and I used the same one.

PR #376, opened 2024-05-13
Java: Throw RuntimeException when a condition isn't met

spinillos:
Contributes to https://github.com/grafana/cog/issues/360

The easiest way to archive this is throwing a `RuntimeException` when a condition isn't met. With this we delegate the exception management to the user.

It also negates all constraints conditions because it was wrong 😅.


julienduchesne:
Can you post a before-after diff of the generated java files? I have trouble reviewing these PRs without examples

testdata/jennies/builders/constraints/JavaBuilders/constraints/SomeStruct.java
```diff
@@ -12,19 +12,19 @@ public Builder() {
             this.internal = new SomeStruct();
         }
     public Builder setId(Long id) {
-        if (id >= 5) {
-            return this;
+        if (!(id >= 5)) {
+            throw new RuntimeException("id must be >= 5");
```
julienduchesne, line 16: nit: These feel more like `IllegalArgumentException`: https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalArgumentException.html which is a subclass of RuntimeException

julienduchesne (approved):

julienduchesne:
Thanks for the diff!

PR #383, opened 2024-05-15
Identifiers formatting wip

K-Phoen:
It's really easy to have inconsistent formatting for identifiers (type names, variables, classes, fields, ...)

This processing is split across various places: different jennies and templates mainly.

What about applying language-specific formatting conventions on the intermediate representations instead? It would guarantee a consistent output, and unclutter jennies+templates.

This PR experiments with this idea.

Would help with #235


K-Phoen:
I still like the idea, but it needs more work.

PR #387, opened 2024-05-20
Generate assignment nil-checks in IR instead of in every jenny

K-Phoen:
To simplify the work needed to introduce/maintain new languages, null-checks performed on assignments should be modeled by the IR.

This allows us to move the fairly complex logic needed to generate these statements somewhere where it can be reused, and remove a bunch of duplicated code from jennies.


internal/jennies/java/templates/types/class.tmpl
```diff
@@ -1,7 +1,11 @@
 package {{ .Package }};
+{{- $class := include "class" . }}
```
spinillos, line 2: I don't understand this change.

K-Phoen: That's because rendering this partial template can trigger the addition of new imports. To be able to include these in the final rendered template, I render this partial into a variable before the imports themselves are rendered.

spinillos (approved):

PR #397, opened 2024-05-22
Add a quick and dirty script to generate a list of configurable compiler passes

K-Phoen:
Rather than having to look at the source to know which compiler passes are usable, we could generate a markdown file to list them.

Warning: the generation process for this doc relies on dark magic :see_no_evil:


Duologic (approved):
I didn't look too deep into the code for this, output seems reasonable.

Issue #143, opened 2023-11-14
Time is parsed to string instead of time.Time


spinillos:
Cue and OpenAPI are parsing times as string for Go when we are expecting `time.Time`.

K-Phoen:
#404 adds support for this in parsers as well as for Go.

More work is required for other languages.

Duologic:
Implemented for some languages, new issues can be opened to implement this for additional languages when necessary.

PR #404, opened 2024-05-24
Simplecue/parse time

K-Phoen:
Relates to #143

For cog to be considered as an option to help with codegen use-cases in https://github.com/grafana/grafana-app-sdk, we need to have at least a basic support of "datetime hinted" strings (especially in Go).

### DIFF
--- a/testdata/jennies/rawtypes/time_hint/GoRawTypes/time_hint/types_gen.go
+++ b/testdata/jennies/rawtypes/time_hint/GoRawTypes/time_hint/types_gen.go
@@ -1,0 +1,8 @@
+package time_hint
+
+type ObjTime time.Time
+
+type ObjWithTimeField struct {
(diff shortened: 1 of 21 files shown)


testdata/jennies/rawtypes/time_hint/GoRawTypes/time_hint/types_gen.go
```diff
@@ -0,0 +1,8 @@
+package time_hint
+
+type ObjTime time.Time
+
+type ObjWithTimeField struct {
+	RegisteredAt time.Time `json:"registeredAt"`
```
spinillos, line 6: `time.Time` import is missing?

K-Phoen: It's added by the post-processor :)

spinillos (approved):

PR #405, opened 2024-05-24
Java: Update paths

spinillos:
It moves all packages/imports under `com.grafana.foundation` all generated files into `src/main/java/com/grafana/foundation` to follow the Java project convention. 

It simplifies the library generation for Maven.

### DIFF
--- a/internal/jennies/java/rawtypes.go
+++ b/internal/jennies/java/rawtypes.go
@@ -13,6 +13,7 @@ import (
 )
 
 const projectPath = "src/main/java/com/grafana/foundation"
+const packagePath = "com.grafana.foundation"
 
 type RawTypes struct {
 	imports *common.DirectImportMap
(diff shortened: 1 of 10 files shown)


malcolmholmes (approved):

internal/jennies/java/rawtypes.go
```diff
@@ -12,6 +12,9 @@ import (
 	"github.com/grafana/cog/internal/tools"
 )
 
+const projectPath = "src/main/java/com/grafana/foundation"
+const packagePath = "com.grafana.foundation"
```
K-Phoen, line 16: Shouldn't these come from the jenny's configuration?

spinillos: To be able to change them for any reason?

K-Phoen: In case cog - which is mostly not grafana-specific - is used in another context, yeah :)

spinillos: I just moved everything into the configuration.

PR #412, opened 2024-05-27
Add an option to skip runtime generation for all languages

K-Phoen:
Mirrors #409 for every other language supported by cog.


spinillos:
In some cases, if you skip runtime generation without skip builders generation will generate code that doesn't compile.

K-Phoen:
Yeah, that option is meant to be used in conjunction with the "types-only" mode.

I can update this PR to document and maybe even refuse to generate builders if this flag is turned on?

spinillos:
Yes, something like that. Because Runtime is necessary for Builders and they cannot be switched off independently.

Issue #408, opened 2024-05-26
Use cog to perform codegen in `grafana-app-sdk`


K-Phoen:
For go types: https://github.com/grafana/grafana-app-sdk/blob/324159db30045da4deb4f204b05f627998e89217/codegen/jennies/gotypes.go#L100-L108 (and other places too)

As well as typescript, instead of cuetsy.

K-Phoen:
A PR is open in `grafana-app-sdk` is open to showcase how a cog integration could look like: https://github.com/grafana/grafana-app-sdk/pull/296

K-Phoen:
Done in https://github.com/grafana/grafana-app-sdk/pull/485

PR #432, opened 2024-06-04
Cue/out of root references

K-Phoen:
Contributes to #408 

`grafana-app-sdk`'s codegen pipeline supports cases where code is generated from a specific path within a cue file, while types relevant for the codegen are defined outside of that path:

```cue
schema: {
  spec: { // ← codegen runs on this
    title: string
    origin: #Origin // ← "out of root" reference
  }
  #Origin: { creator: string } 
}
```

### DIFF
--- a/internal/simplecue/utils.go
+++ b/internal/simplecue/utils.go
@@ -216,6 +216,10 @@ func commentsFromCueValue(v cue.Value) []string {
 		}
 	}
 
+	if len(docs) == 0 {
+		return nil
+	}
+
 	ret := make([]string, 0, len(docs))
 	for _, cg := range docs {
 		ret = append(ret, strings.Split(strings.Trim(cg.Text(), "\n "), "\n")...)
--- a/internal/simplecue/utils.go
+++ b/internal/simplecue/utils.go
@@ -272,3 +272,31 @@ func hasStructFields(v cue.Value) bool {
 
 	return false
 }
+
+func areCuePathsFromSameRoot(a cue.Path, b cue.Path) bool {
+	selectorsA := a.Selectors()
+	selectorsB := b.Selectors()
+
+	if len(selectorsA) == 0 || len(selectorsB) == 0 {
+		return false
+	}
+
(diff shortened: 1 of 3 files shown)


internal/simplecue/utils.go
```diff
@@ -272,3 +276,31 @@ func hasStructFields(v cue.Value) bool {
...
+	}
+
+	return selectorsA[0].String() == selectorsB[0].String()
+}
+
+func cuePathIsChildOf(root cue.Path, maybeChild cue.Path) bool {
+	selectorsRoot := root.Selectors()
+	selectorsChild := maybeChild.Selectors()
+
+	if len(selectorsRoot) > len(selectorsChild) {
```
spinillos, line 295: Could happen that child has more selectors than the root?

K-Phoen: Yeah :)

```cue
schema: {
  spec: { // path: schema.spec (2 selectors)
    title: string
    origin: #Origin
  }
  #Container: {
    foo: string
    #OtherContainer: {
      bar: string

      #Origin: { creator: string } // path: schema.#Container.#OtherContainer.#Origin (4 selectors)
    }
  }
}
```

spinillos: And in the next loop you are checking if any field is different to return false, but if `selectorsChild` is bigger, you aren't checking all the fields, right? Or am I missing something?

K-Phoen: Yeah, the idea is to check if one selector is a "subset" of the other.

Let's imagine:
* `root = schema.spec`
* `maybeChild = schema.spec.#Container.#Origin`
* in that case, we are not in an "out of root" scenario as `root` is a subset of `maybeChild`

Whereas in this example, we do have an "out of root" reference:
* `root = schema.spec`
* `maybeChild = schema.#Container.#Origin`

In both cases, we need to check the selectors of `root` to make a decision but we might not have to check all the selectors of `maybeChild`

spinillos (approved):

PR #433, opened 2024-06-05
Java: Publishing to foundation-sdk

spinillos:

_It shouldn't be merged until we prepare some examples and a few more tests to ensure that its working as expected_.

This PR creates the CI steps to be able to do the checks to the Java code and the publishing step.

Following the same way that other teams doing the publishing in Grafana, we removed the pom.xml file in favor of gradle file. Example: https://github.com/grafana/grafana-opentelemetry-starter.

It needs to setup some credentials in `foundation-sdk` settings to make it works.


julienduchesne (approved):
:shipit:

PR #441, opened 2024-06-10
Java: Add builder's implementation

spinillos:
It adds builder's implementation to Builders to be able to use in functions.


internal/jennies/java/builders.go
```diff
@@ -52,6 +52,7 @@ func (b Builders) genBuilder(pkg string, name string) (template.Builder, bool) {
 		Options:     builder.Options,
 		Properties:  builder.Properties,
 		Defaults:    b.genDefaults(builder.Options),
+		ImportAlias: b.typeFormatter.formatPackage("cog.Builder"),
```
spinillos, line 55: I used import alias to avoid to create a new object only to add this.

testdata/jennies/builders/anonymous_struct/JavaBuilders/anonymous_struct/SomeStruct.java
```diff
@@ -4,7 +4,7 @@
 public class SomeStruct {
     public Object time;
     
-    public static class Builder {
+    public static class Builder implements cog.Builder<SomeStruct> {
```
julienduchesne, line 7: Should `cog` be imported? Am I misunderstanding something about Java?

K-Phoen (approved):

PR #447, opened 2024-06-11
Java: Use schema object name to create the map of builders

spinillos:
In Java we have a map of builders used to decide if we should add the builder or not inside the class (common way to do in Java).

The problem is we were iterating the builder structure only and the names of the Builders differs from the one in Schema and it was impossible to find them.

### DIFF
--- a/internal/jennies/java/builders.go
+++ b/internal/jennies/java/builders.go
@@ -25,8 +25,7 @@ func parseBuilders(config Config, context languages.Context, formatter *typeForm
 			b[builder.Package] = map[string]ast.Builder{}
 		}
 
-		obj, _ := context.LocateObject(builder.For.SelfRef.ReferredPkg, builder.For.SelfRef.ReferredType)
-		b[builder.Package][obj.Name] = builder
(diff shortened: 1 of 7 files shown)


internal/jennies/java/builders.go
```diff
@@ -24,7 +24,9 @@ func parseBuilders(config Config, context languages.Context, formatter *typeForm
 		if _, ok := b[builder.Package]; !ok {
 			b[builder.Package] = map[string]ast.Builder{}
 		}
-		b[builder.Package][builder.Name] = builder
+
+		obj, _ := context.LocateObject(builder.For.SelfRef.ReferredPkg, builder.For.SelfRef.ReferredType)
+		b[builder.Package][obj.Name] = builder
```
K-Phoen, line 29: If you only need to access the object's name, you don't need to locate it :)

This should be equivalent:
```suggestion
		b[builder.Package][builder.For.SelfRef.ReferredType] = builder
```

spinillos: Right!

K-Phoen (approved):

PR #457, opened 2024-06-14
Java: Fix multiple constructors names

spinillos:
Constructor name and function returns should use the class name.


K-Phoen (approved):
I knew I'd break something with my PR :sweat_smile: 

Time to add Java linting to our CI it seems!

PR #472, opened 2024-06-20
Compiler: Fix testdata datasource name

spinillos:
All renaming names are in upper camel case except this one. It generates wrong Java code.


julienduchesne (approved):

K-Phoen (approved):
Shouldn't matter much anyway: jennies will format this :)

PR #474, opened 2024-06-20
Fix test data for python/openapi

malcolmholmes:
Regenerating the test data causes these two minor changes. This makes test data regeneration a no-op.﻿


spinillos (approved):

K-Phoen:
Tests data have been regenerated a few times since this PR, so it shouldn't be needed anymore.

PR #478, opened 2024-06-21
Java: Example

spinillos:
Java example


examples/java/CPU.java
```diff
@@ -0,0 +1,80 @@
...
+
+import java.util.List;
+
+public class CPU {
+    public static Builder<Panel> cpuUsageTimeseries() {
+        String query = "((1 - sum without (mode) (rate(node_cpu_seconds_total{job=\"integrations/raspberrypi-node\", mode=~\"idle|iowait|steal\", instance=\"$instance\"}[$__rate_interval]))) " +
+                "/ ignoring(cpu) group_left " +
+                "count without (cpu, mode) (node_cpu_seconds_total{job=\"integrations/raspberrypi-node\", mode=\"idle\", instance=\"$instance\"}))";
+
+        Threshold th1 = new Threshold();
```
K-Phoen, line 17: We should provide constructors for classes (not required now, an issue *for later*™ is fine)

examples/java/Main.java
```diff
@@ -0,0 +1,97 @@
...
+                WithRow(new RowPanel.Builder("Network")).
+                WithPanel(Network.networkReceivedTimeseries()).
+                WithPanel(Network.networkTransmittedTimeseries()).
+                // Logs
+                WithRow(new RowPanel.Builder("Logs")).
+                WithPanel(Logs.errorsInSystemLogs()).
+                WithPanel(Logs.authLogs()).
+                WithPanel(Logs.kernelLogs()).
+                WithPanel(Logs.allSystemLogs()).
+                build();
```
K-Phoen, line 45: I'm a bit surprised to see the `build()` method in lowercase while other methods seem to all be UpperCamelCase. Is that on purpose? :thinking:

K-Phoen (approved):
What's the sdk-1.0.jar file? :thinking:

examples/java/Main.java
spinillos: Nop, its a mistake. Sometimes I get confused with Go 😓

Issue #470, opened 2024-06-18
PHP support


K-Phoen:
Types, builders, ... the whole thing.

PR #486, opened 2024-06-24
Language/php examples

K-Phoen:
This PR provides a better example for PHP dashboard builders, and runs it in the CI.
A few tiny fixes were also needed along the way. 

Contributes to #470


spinillos (commented):
You fixed a pending issue with the safe guards too :)

spinillos (approved):

K-Phoen:
> You fixed a pending issue with the safe guards too :)

Did I? :sweat_smile:

PR #487, opened 2024-06-25
Fix constants generation in PHP

K-Phoen:
In PHP, constants can't be autoloaded if they're not defined within a class.

Contributes to #470


julienduchesne (approved):
PHP is weird as hell

K-Phoen:
That it definitely can be :D

PR #489, opened 2024-06-25
Update the foundation-sdk maturity statement to "public preview"

K-Phoen:
Let's update the maturity statement displayed in the foundation-sdk to something less scary, reflecting the actual state of the sdk.


julienduchesne (approved):
👍

PR #490, opened 2024-06-25
Language/php foundation sdk readme

K-Phoen:
Contributes to #470


julienduchesne (approved):
Not very good at PHP, but the format looks good

PR #494, opened 2024-06-27
Java: Marshalling

spinillos:
It fixes the marshalling for disjunctions (refs and scalars). For example, when you add a panel or a row inside panel's array in the dashboard, you are expecting to have the internal structure of a panel or row. In Java we were adding the whole disjunction struct adding extra values.

The expected behaviour is:
```json
{
  "panels": [{ "type": "row" }, { "type": "panel" }]
}
```

And in Java we had:
```json
"panels": [{"row": { "type": "row" }}, {"panel": { "type": "panel" }}]
```

And a Java example to see better what's the code is doing:
### Before
```java
public class StringOrMap {

    @JsonProperty("string")
    public String string;
    @JsonProperty("map")
    public Map<String, Object> map;

    public String toJSON() throws JsonProcessingException {
        ObjectWriter ow = new ObjectMapper().writer().withDefaultPrettyPrinter();
        return ow.writeValueAsString(this);
    }
}
```

### After
```java
public class StringOrMap {

    @JsonUnwrapped
    public String string;
    @JsonUnwrapped
    public Map<String, Object> map;

    public String ToJSON() throws JsonProcessingException {
        if (string != null) {
            ObjectWriter ow = new ObjectMapper().writer().withDefaultPrettyPrinter();
            return ow.writeValueAsString(string);
        }
        if (map != null) {
            ObjectWriter ow = new ObjectMapper().writer().withDefaultPrettyPrinter();
            return ow.writeValueAsString(map);
        }

        return null;
    }
}
```

### DIFF
--- a/internal/jennies/java/jsonmarshalling.go
+++ b/internal/jennies/java/jsonmarshalling.go
@@ -1,0 +1,22 @@
+package java
+
+import (
+	"strings"
+
+	"github.com/grafana/cog/internal/ast"
+)
+
+func genToJSONFunction(t ast.Type) string {
+	var buffer strings.Builder
+	if t.IsStructGeneratedFromDisjunction() {
+		if t.IsStruct() && (t.HasHint(ast.HintDiscriminatedDisjunctionOfRefs) || t.HasHint(ast.HintDisjunctionOfScalars)) {
+			_ = templates.ExecuteTemplate(&buffer, "marshalling/disjunctions.json_marshall.tmpl", map[string]any{
+				"Fields": t.AsStruct().Fields,
+			})
+			return buffer.String()
+		}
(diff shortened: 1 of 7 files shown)


K-Phoen (approved):
I'm surprised to see no changes to testdata files. Are we missing test cases for this type of disjunction? :thinking:

spinillos:
@K-Phoen Its because the json stuff is using a dependency that isn't going to work if we don't generate all gradle/maven on top of that. 

The options could be generating the whole gradle/maven project by default to ensure that we are generating valid/compiling code or omit all json related. Without using the dependency, the JSON generated using only the models probably is going to be wrong. I'm open to any idea about this.

PR #499, opened 2024-07-03
Java: Register and deserialisers

spinillos:
This PR adds panels, dataquery and disjunctions deserialiser. Also it adds missing changes to make it works:
* It creates a registry like [Go](https://github.com/grafana/grafana-foundation-sdk/blob/next%2Bcog-v0.0.x/go/cog/plugins/variants.go) one that registers dataqueries and panels. It allows to add custom panels/dataqueries ([Code](https://github.com/grafana/cog/pull/499/files#diff-853abaa5badae5b5fb681cb25e05fa6aa7b91f5dc2e34063897f258da59eb03fR1)).
* Classes that need a specific deserialisers adds a annotation on top of them pointing to their deserialiser. ([Code](https://github.com/grafana/cog/pull/499/files#diff-e139447fb637f49c01aee0a075ccd82c57a91271e252739addac7614220ee3d4R13-R15))
* There is a specific serialiser for disjunctions of refs and scalars similar to Go ([Code](https://github.com/grafana/cog/pull/499/files#diff-42ec7b5563f875e24518da3b1350d86ce477648ddded143f23c251681356e709R1), [Code](https://github.com/grafana/cog/pull/499/files#diff-690f47704106561b95af8d973f70907ee87c246229a2fa8163091595fe1d2c02R1)).
* Panels has their own one to be able to set the proper panel options (and optionally fieldConfig). ([Code](https://github.com/grafana/cog/pull/499/files#diff-25ba755beae2fd0c3dea2d149496933ce2165837f7d5d845d0ff12f9d06fdbabR41-R64))
* The ones that uses dataqueries internally are able to map to the specific dataquery. ([Code](https://github.com/grafana/cog/pull/499/files#diff-25ba755beae2fd0c3dea2d149496933ce2165837f7d5d845d0ff12f9d06fdbabR67-R110))
  * It adds EmptyDataquery class that its used to map unknown dataqueries. ([Code](https://github.com/grafana/cog/pull/499/files#diff-25ba755beae2fd0c3dea2d149496933ce2165837f7d5d845d0ff12f9d06fdbabR99-R104))
* It adds empty case in enums to avoid to crash when we deserialise a empty enum value. ([Code](https://github.com/grafana/cog/pull/499/files#diff-f0a7ceb33be772e3591df7fe113e5d729793d16dde0e5cd89e171c457568941cR194-R208))


internal/jennies/java/deserializers.go
```diff
@@ -0,0 +1,164 @@
...
+
+func (jenny *Deserializers) genDataqueryCode(context languages.Context, obj ast.Object) (DataqueryUnmarshalling, error) {
+	for _, field := range obj.Type.AsStruct().Fields {
+		composableSlotType, resolved := context.ResolveToComposableSlot(field.Type)
+		if !resolved {
+			continue
+		}
+
+		if composableSlotType.AsComposableSlot().Variant == ast.SchemaVariantDataQuery {
+			return jenny.renderUnmarshalDataqueryField(obj, field), nil
```
malcolmholmes, line 90: return or continue? Is there only one composable slot per struct?

malcolmholmes (approved):

internal/jennies/java/registry.go
```diff
@@ -0,0 +1,142 @@
...
+			continue
+		}
+
+		if schema.Metadata.Variant == ast.SchemaVariantDataQuery {
+			dataquerySchemas = append(dataquerySchemas, DataquerySchema{
+				Identifier: strings.ToLower(schema.Metadata.Identifier),
+				Class:      jenny.formatPackage(fmt.Sprintf("%s.%s", schema.Package, jenny.findDataqueryClass(schema))),
+			})
+		} else {
+			panelSchemas = append(panelSchemas, PanelSchema{
```
Duologic, line 74: Shouldn't this also test for Panel variant and fail/continue if it's not?

internal/jennies/java/deserializers.go
spinillos: It should be a continue, right. It was working because we don't have more than one case right now 😅

PR #500, opened 2024-07-04
Java: Move to gradle

spinillos:
It generates Java code as a Gradle project directly without any extra flag in configurations.

Why?? There are several reasons:

* Gradle is used in Grafana projects instead of Maven.
* Its easier to verify that the code compiles if we do it using Gradle or Maven, since we need to add external dependencies.
* We avoid juggling to avoid to break the CI every time that we change something in Java, spending hours to make it works and avoid to want to kill ourself or get depressed. (example: https://github.com/grafana/cog/pull/499)
* Tests reflects better what we are doing.
* We are able to execute Java examples in CI (not yet implemented).
* Adding Gradle files into the generated code don't add extra overhead.


malcolmholmes:
I see changes to (a) switch to Gradle and (b) change how types are annotated/marshalled. Are they both a part of the same PR?? Does the first depend on the second?

spinillos:
@malcolmholmes Yes because the pom generation and maven version were part of Maven configuration. Previously was optional and I set them mandatory to we able to reflect it into tests.

I did a small refactor to remove some flags because we don't need them anymore.

malcolmholmes (approved):

PR #510, opened 2024-07-16
Java: Fix release templates

spinillos:
Missing `actions/checkout@v4` step and `gradle build` should be a different step.

### DIFF
--- a/repository_templates/java/.github/workflows/java-ci.yaml
+++ b/repository_templates/java/.github/workflows/java-ci.yaml
@@ -27,8 +27,6 @@ jobs:
 
       - name: Java setup Gradle
         uses: gradle/actions/setup-gradle@v3
-        env:
-          CHECK_GENERATED_FILES: true
         with:
           gradle-version: '8.8'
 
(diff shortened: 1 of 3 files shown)


repository_templates/java/.github/workflows/java-ci.yaml
```diff
@@ -16,17 +16,21 @@ jobs:
...
 
-      - name: Test and build Jar
+      - name: Java setup Gradle
         uses: gradle/actions/setup-gradle@v3
         env:
           CHECK_GENERATED_FILES: true
         with:
           gradle-version: '8.8'
+
+      - name: Java build
```
julienduchesne, line 35: The env. CHECK_GENERATED_FILES should be in this step also, right?

spinillos: I was checking this env variable and I saw that it isn't used in our project. I followed how it was done in other Grafana project but it has a lot of things deprecated.

I also updated the release CI workflow.

julienduchesne (approved):

Issue #513, opened 2024-08-01
Java: Add examples in foundation-sdk


spinillos:
Java isn't added into the documentation and we miss some examples. There is a started [PR](https://github.com/grafana/grafana-foundation-sdk/tree/java/examples) but not finished.

PR #524, opened 2024-08-29
Java: Create constructors for models without them

spinillos:
Closes https://github.com/grafana/cog/issues/513

It adds empty constructors and the ones with all fields. The empty one exists by default when we don't have any, but it stops to exist if we decide to create one. So its why we generate it.


spinillos:
Closing it since we need to add them only to specific Objects.

Issue #534, opened 2024-09-06
Avoid to update version in README files when it doesn't publish a new release


spinillos:
The release workflows for Java, Python and Typescript don't happen when the code doesn't detect any change in their files. But at the same time we are updating the readme or some files with the new timestamp for the new release.

So we have two options:
* Don't update these files.
* Release even if we don't have changes. (In that case I think that we should remove `paths`).

K-Phoen:
This should have been resolved by #545 :)

PR #545, opened 2024-09-12
Always trigger a release in foundation-sdk when a package file is modified

K-Phoen:
Should help with #534


scripts/release-version.sh
```diff
@@ -57,10 +57,6 @@ function run_codegen() {
   local extra_parameters="${1}"
   shift
 
-  echo $COG_CMD generate \
```
K-Phoen, line 60: A stray debug statement made its way here :scream:

Issue #314, opened 2024-04-09
Support converting JSON to resources?


K-Phoen:
Kind of the opposite from what a builder does (code → JSON), it would be a useful improvement in terms of DX to be able to convert JSON to the "builder code" representing it.

How feasible is this feature? Could it span across languages? (ie: a Go _converter library_ would be able to produce Go, TS, Python, ...)

PR #566, opened 2024-09-27
Bootstrap JSON → code conversion in Go

K-Phoen:
Contributes to #314 

This PR lays the foundation to convert JSON to code, starting with an implementation in Go.
There are still more things to do:

* [ ] tests
* [x] take advantage of #560 to more efficiently unmarshal (and convert) dataqueries
* [ ] implement for other languages (the current conversion process works for Go, but it will most likely need some tweaks to fully support other languages... disjunctions, I'm especially looking at you)
* [x] Options with no arguments aren't converted?
	* `Editable()` on dashboard
	* `Instant()` on prometheus queries
	* ...
* [x] skip options whose value is the default one: kinda done → for scalars only (simplest case to detect)

### DIFF
--- a/internal/ast/types.go
+++ b/internal/ast/types.go
@@ -514,6 +514,10 @@ func NewObject(pkg string, name string, objectType Type, passesTrail ...string) 
 	}
 }
 
+func (object *Object) AsRef() Type {
+	return object.SelfRef.AsType()
+}
+
 func (object *Object) AddToPassesTrail(trail string) {
 	object.PassesTrail = append(object.PassesTrail, trail)
 }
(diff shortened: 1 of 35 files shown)


internal/ast/types.go
```diff
@@ -514,6 +514,10 @@ func NewObject(pkg string, name string, objectType Type, passesTrail ...string)
 	}
 }
 
+func (object *Object) AsRef() Type {
```
spinillos, line 517: `AsRef` looks like that I'm expecting a `Ref`, not the `Type`.

spinillos (approved):

PR #578, opened 2024-10-06
[PoC] grizzly resources in Go

K-Phoen:
The goal of this PR is to experiment with a way to integrate a bit better the resources generated by the foundation-sdk with grizzly.

With that in mind, it would be useful for some SDK resources to be generated in a format directly compatible with grizzly.

To achieve this, I added two things to the Go jennies:
* `yaml` tags to structs, describing how to marshal them into YAML (one of the formats supported by grizzly). Custom `MarshalYAML()` methods are also generated for disjunctions. Note: unmarshalling is left un-handled (for now?)
* the ability to add arbitrary methods to generated Go structs via templates configurable at runtime

This allows us to write the following:

```go
dashboardBuilder := /* lala */
dashboard, err := dashboardBuilder.Build()
if err != nil {
	panic(err)
}

dashboardYaml, err := yaml.Marshal(dashboard.AsGrizzlyResource())
if err != nil {
	panic(err)
}

if err := os.WriteFile("dashboard.yaml", dashboardYaml, 0644); err != nil {
	panic(err)
}
```

And then grizzly can take over, pushing the dashboard/serving it locally/...

Other objects supported by Grizzly could benefit from the same customization.
Other languages could be updated to support similar features.

Thoughts?


spinillos (approved):

spinillos (changes requested):
We need to talk about if we want to have grizzly specific stuff into foundation-sdk

K-Phoen:
Closing this for now, let's pick it up again later if it's still relevant :)

PR #583, opened 2024-10-07
Support choices between builders for a given type

K-Phoen:
Contributes to #314 

This PR fixes the converter logic for types mapped to several builders.

It allows the converter to generate code for the relevant builder, depending on the provided input.

More concrete example: https://github.com/grafana/cog/blob/eff72aecd692de946caf1872c095381224f69ac4/config/veneers/dashboard.variables.common.yaml#L15

Dashboard variables have several builders (one for variable type), while the underlying type always is `VariableModel` (since the schema defines a *catch-all* type for all variables)

### DIFF
--- a/internal/ast/builder.go
+++ b/internal/ast/builder.go
@@ -157,6 +157,7 @@ func (builders Builders) HaveConstantConstructorAssignment() bool {
 		for _, assignment := range builder.Constructor.Assignments {
 			if assignment.HasConstantValue() {
 				constantAssignmentFound = true
+				break
 			}
 		}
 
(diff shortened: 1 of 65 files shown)


internal/ast/builder.go
```diff
@@ -140,6 +140,34 @@ func (builders Builders) LocateByObject(pkg string, name string) (Builder, bool)
...
+
+	return results
+}
+
+func (builders Builders) HaveConstantConstructorAssignment() bool {
+	for _, builder := range builders {
+		constantAssignmentFound := false
+		for _, assignment := range builder.Constructor.Assignments {
+			if assignment.HasConstantValue() {
+				constantAssignmentFound = true
```
spinillos, line 159: break here?

spinillos (approved):

PR #595, opened 2024-10-11
Add a cli tool to the Go version of foundation-sdk

K-Phoen:


spinillos (approved):

K-Phoen:
I'll wait to merge this until we can do a proper validation of resources, instead of the naive "json unmarshal" that we currently do.

package_templates/go/cmd/cli/validate/commands.go
```diff
@@ -0,0 +1,140 @@
...
+	panel := dashboard.Panel{}
+	if err := panel.UnmarshalJSONStrict(input); err != nil {
+		return nil, err
+	}
+
+	return panel, nil
+}
+
+func unmarshalQueryStrict(input []byte, hint string) (validableResource, error) {
+	query, err := cog.StrictUnmarshalDataquery(input, hint)
```
spinillos, line 134: `StrictUnmarshalDataquery` should be `UnmarshalJSONStrict` like dashboards and panels to follow the same pattern. It could be fixed in other PR.

spinillos (approved):

package_templates/go/cmd/cli/validate/commands.go
K-Phoen: Good catch! I'll fix that :+1:

K-Phoen:
Let's wait to merge this: I feel we need to properly think about how to expose CLI capabilities (and if it makes sense at all)

PR #598, opened 2024-10-15
Ignore vendor and Java folders

spinillos:
When you generate the sdk or execute Java tests in local, it generates some build folders that we should ignore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,5 +3,10 @@
 /examples/php/vendor
 /.devbox
 /gradle.properties
+/vendor
+
+# Java building folders
+.gradle/
+build/
 
 .idea


.gitignore
```diff
@@ -3,5 +3,10 @@
 /examples/php/vendor
 /.devbox
 /gradle.properties
+/vendor
```
K-Phoen, line 6: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/ :trollface:

K-Phoen (approved):

.gitignore
spinillos: I have it for .DS_Store and a few things more. I thought that vendor wasn't one of them.

PR #599, opened 2024-10-16
Go/strict json unmarshal

K-Phoen:
This PR adds the capacity for go types to be "strictly unmarshalled".

### DIFF
--- a/internal/jennies/golang/strictjson.go
+++ b/internal/jennies/golang/strictjson.go
@@ -12,15 +12,13 @@ import (
 
 type strictJSONUnmarshal struct {
 	tmpl          *template.Template
-	config        Config
 	imports       *common.DirectImportMap
 	packageMapper func(string) string
 	typeFormatter *typeFormatter
 }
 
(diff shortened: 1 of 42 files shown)


internal/jennies/golang/strictjson.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (jenny strictJSONUnmarshal) objectNeedsUnmarshal(obj ast.Object) bool {
+	return obj.Type.IsStruct()
+}
+
+func (jenny strictJSONUnmarshal) renderUnmarshal(context languages.Context, obj ast.Object) (string, error) {
+	tmpl := jenny.tmpl.
+		Funcs(common.TypeResolvingTemplateHelpers(context)).
+		Funcs(template.FuncMap{
+			"resolvesToArrayOfScalars": func(typeDef ast.Type) bool {
```
spinillos, line 59: Could be move these ones into the common helpers?

spinillos (approved):

internal/jennies/golang/strictjson.go
K-Phoen: Not really: I cheated and included enums in what these functions consider to be scalars (which is valid in Go but not necessarily in other languages)

PR #601, opened 2024-10-18
Cache go vendors in CI

K-Phoen:
Take advantage of build/result caches to speedup the CI.

Devbox is still the most time-consuming task, not sure how to address it right now but it also installs an absolute bunch of things so... :shrug:


spinillos (commented):
Devbox has a [cache configuration](https://github.com/marketplace/actions/devbox-installer#action-inputs) that its `false` by default. Maybe it could help.

K-Phoen:
We've already enabled it :)

But we install so many languages/tools that even just downloading and extracting the cache takes a bunch of time.

I'll take a look at how we could maybe speed that up *later* ©

spinillos (approved):

Issue #512, opened 2024-08-01
Java: Types without builder should ask fields as constructor arguments


spinillos:
We have some veneers to skip builder's generation for some types that only contains one or two fields. In that case we need to generate the object initialising its struct directly.

In case of Java, you cannot initialise the struct directly like other languages. You need to initialise the object and after that setting their values after that. It is ugly since current generation makes everything public and set the values directly is not the "Java way".

So to simplify this pain, we could ask for the values into their constructor. In Java you can overload the constructors and methods, so probably the best way to cover it is to generate all these constructors.

For example:
```java
public class MyClass {
   public String a;
   public Integer b;
}
```
should have:
```java
public MyClass(String a, Integer b) {
  this.a = a;
  this.b = b;
}

public MyClass(String a) {
  this.a = a;
}

public MyClass(Integer b) {
  this.b = b;
}
```

If its complex, at least with the first constructor we can make it works. We are using objects for primitive types so `null` value is acceptable as argument.

**If I'm not wrong, it can be fixed using veneers.**

PR #611, opened 2024-10-25
Java: Constructors for no builders

spinillos:
Closes: https://github.com/grafana/cog/issues/512

It adds constructors for [these](https://github.com/grafana/cog/blob/main/config/veneers/dashboard.common.yaml#L13-L21) schemas without builder. We can omit some cases where we are creating the empty object and set values one to one. 
Apart of this, it simplifies to set the values of the converters to avoid to create extra code to generate an object.

[These](https://github.com/grafana/cog/blob/main/config/veneers/dashboard.common.yaml#L82-L83) other two are skipped because builders asks for their internal values, but not for the whole object.

### DIFF
--- a/examples/java/src/main/java/test/CPU.java
+++ b/examples/java/src/main/java/test/CPU.java
@@ -16,38 +16,32 @@ public class CPU {
                 "/ ignoring(cpu) group_left " +
                 "count without (cpu, mode) (node_cpu_seconds_total{job=\"integrations/raspberrypi-node\", mode=\"idle\", instance=\"$instance\"}))";
 
-        Threshold th1 = new Threshold();
-        th1.color = "green";
-
-        Threshold th2 = new Threshold();
-        th2.color = "red";
-        th2.value = 80.0;
-
         return Common.defaultTimeSeries().
                 title("CPU Usage").
                 span(18).
                 stacking(new StackingConfig.Builder().mode(StackingMode.NORMAL)).
                 thresholds(new ThresholdsConfig.Builder().
                         mode(ThresholdsMode.ABSOLUTE).
-                        steps(List.of(th1, th2))).
+                        steps(List.of(
+                                new Threshold(0.0, "green"),
+                                new Threshold(80.0, "red")
+                        ))
+                ).
                 max(1.0).
                 min(0.0).
                 withTarget(Common.basicPrometheusQuery(query, "{{ cpu }}"));
     }
 
     public static Builder<Panel> loadAverageTimeseries() {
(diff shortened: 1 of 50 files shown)


examples/java/src/main/java/test/CPU.java
```diff
@@ -16,38 +16,32 @@ public static Builder<Panel> cpuUsageTimeseries() {
...
-
         return Common.defaultTimeSeries().
                 title("CPU Usage").
                 span(18).
                 stacking(new StackingConfig.Builder().mode(StackingMode.NORMAL)).
                 thresholds(new ThresholdsConfig.Builder().
                         mode(ThresholdsMode.ABSOLUTE).
-                        steps(List.of(th1, th2))).
+                        steps(List.of(
+                                new Threshold(0.0, "green"),
```
K-Phoen, line 26: Much better!

internal/jennies/java/rawtypes.go
```diff
@@ -298,3 +299,30 @@ func (jenny RawTypes) getVariant(t ast.Type) string {
 	}
 	return variant
 }
+
+func (jenny RawTypes) defaultConstructor(object ast.Object) []ast.Argument {
```
K-Phoen, line 303: Why not generating constructors for every object?

internal/jennies/java/templates/types/class.tmpl
```diff
@@ -28,6 +28,14 @@ public class {{ .Name }}{{ if .Extends }} extends {{ range $i, $e := .Extends }}
         return {{ $.Name | lowerCamelCase }};
     }
     {{- end }}
+    {{- else if .DefaultConstructorArgs }}
+    public {{ $.Name }}() {}
```
K-Phoen, line 32: What is this empty constructor useful for?
If some fields in the object are required, it would allow constructing an object in an "invalid" state

K-Phoen (approved):
Some questions, but nothing critical :)

internal/jennies/java/templates/types/class.tmpl
spinillos: When you create a constructor, the default disappears unless you define it. I added to avoid to refactor the builder's code to use the new ones. We could eventually do it to make the code a bit nicer.

internal/jennies/java/rawtypes.go
spinillos: Because it generates monster builders that they aren't used in builders... So at the moment I didn't want to add them. I preferred to add them mainly to the builders that we omit in veneers because they have a few fields only.

PR #612, opened 2024-10-25
Java: Converters

spinillos:
Contributes to https://github.com/grafana/cog/issues/591

It creates converters for Java. Converter is commented on jenny to avoid to publish it since it needs a bit more work...

* It creates `Runtime.go` for all functions related of how to parse generics, enums and some cleanup functions for strings.
  * It detects string blocks and it sets them with three double quotes. With that it doesn't need to escape the text and simplify it. 
* Adds Converter interface to identify them.
* In Java everything is an object and they could be nullable. So when an object isn't nullable in the schema, it doesn't check this and its why that in `guard` block in `converter.tmpl` is checking this to add the missing check.

It misses:
* Manage unknown panels and dataqueries
* Add missing constructor parameters and functions for generic objects. At the moment its assuming that constructors don't have parameters.
* Tests 🌟


internal/jennies/java/converter.go
```diff
@@ -0,0 +1,167 @@
...
+	packageMapper := func(pkg string, class string) string {
+		if imports.IsIdentical(pkg, schema.Package) {
+			return ""
+		}
+
+		return imports.Add(class, pkg)
+	}
+	typeFormatter := createFormatter(context, jenny.config).withPackageMapper(packageMapper)
+
+	builderNameFormat := fmt.Sprintf("%s.Builder", tools.UpperCamelCase(converter.BuilderName))
```
spinillos, line 89: There are some builders that the name of the class is different than the builder because they are renamed by a veneer. Apart of this, we have classes with different constructors depending of they type (`VariableModel` for example), and the builder name is also different.

internal/jennies/java/templates/converters/converter.tmpl
```diff
@@ -0,0 +1,181 @@
...
+{{- .Imports }}
+{{ $converter }}
+
+{{- define "guard" }}
+    {{- $leftOperand := print (.Path | formatGuardPath) -}}
+    {{- $type := .Path.Last.Type }}
+    {{- if and (eq .Op "!=") (eq .Value nil) -}}
+        {{ $leftOperand }} != null
+    {{- else if and $type.IsScalar (eq $type.AsScalar.ScalarKind "bool") -}}
+        {{- if not $type.Nullable }}
```
spinillos, line 13: This is the problem that I mentioned in Java for no `nullables`. There is only one guard in that cases and it only check the emptiness of the field when we also need the nullability.

internal/jennies/java/templates/converters/converter.tmpl
```diff
@@ -0,0 +1,181 @@
...
+        {{- if and (eq .Type.AsScalar.ScalarKind "string") -}}
+        {{- $reflect := importStdPkg "cog" "Runtime" -}}
+        Runtime.escapeString({{- .Path | formatGuardPath }})
+        {{- else }}
+        {{- .Path | formatGuardPath }} + "{{ formatIntegerLetter .Type }}"
+        {{- end }}
+    {{- else if resolvesToEnum .Type -}}
+        {{- $reflect := importStdPkg "cog" "Runtime" -}}
+        {{- $reflect := importStdPkg .Type.AsRef.ReferredPkg .Type.AsRef.ReferredType -}}
+        Runtime.formatEnum({{ .Type.AsRef.ReferredType }}.class, {{- .Path | formatGuardPath }})
```
spinillos, line 61: For enums we need to know the class because if the value is null, we cannot know the class. In Java we cannot use the string or the integer directly. We need to set `MyEnumClass.DEFAULT_VALUE`.

internal/jennies/java/templates/converters/converter_runtime.tmpl
```diff
@@ -0,0 +1,147 @@
...
+        }
+
+        if (object.getClass().getConstructors().length == 0) {
+            return formatFactory(object);
+        }
+
+        return formatConstructor(object);
+    }
+    
+    public static String formatFactory(Object object) {
```
spinillos, line 31: Don't ask...

internal/jennies/java/templates/converters/converter_runtime.tmpl
```diff
@@ -0,0 +1,147 @@
...
+
+        return sb.toString();
+    }
+
+    public static String fieldAsFunctionName(String f) {
+        return "create"  + f.substring(0, 1).toUpperCase() + f.substring(1);
+    }
+    
+    public static String escapeString(String input) {
+        if (input.contains("\n")) {
```
spinillos, line 101: Strings with new lines, we set them as blocks to avoid to escape characters.

internal/jennies/java/converter.go
```diff
@@ -0,0 +1,167 @@
...
+			formatPackageName(builder.Package),
+			fmt.Sprintf("%sConverter.java", tools.UpperCamelCase(builder.Name)),
+		)
+
+		files = append(files, *codejen.NewFile(filename, output, jenny))
+	}
+
+	var err error
+	for _, schema := range context.Schemas {
+		schema.Objects = schema.Objects.Filter(func(key string, obj ast.Object) bool {
```
K-Phoen, line 46: I don't think we should assign the result of filtering directly into `schema.Objects`: this updates the schema and if another jenny runs after this one, it will be missing some objects

internal/jennies/java/converter.go
```diff
@@ -0,0 +1,167 @@
...
+		return imports.Add(class, pkg)
+	}
+	typeFormatter := createFormatter(context, jenny.config).withPackageMapper(packageMapper)
+
+	builderNameFormat := fmt.Sprintf("%s.Builder", tools.UpperCamelCase(converter.BuilderName))
+	if isPanel {
+		builderNameFormat = fmt.Sprintf("%s.PanelBuilder", typeFormatter.formatPackage(converter.Package))
+	} else if converter.BuilderName != converter.Input.TypeRef.ReferredType {
+		isRenamed := false
+		for _, v := range builder.VeneerTrail {
```
K-Phoen, line 94: That's very very hacky :scream: (and will only work if `debug` is on)

We need to find another way to build `builderNameFormat`

spinillos: Maybe I can find other way to set the names of the builders. This solution is horrible.

internal/jennies/java/converter.go
spinillos: Gotcha!. I though that it only affects to this jenny.

internal/jennies/common/templatehelpers.go
```diff
@@ -27,6 +27,10 @@ func TypeResolvingTemplateHelpers(context languages.Context) template.FuncMap {
 			_, found := context.ResolveToComposableSlot(typeDef)
 			return found
 		},
+		"resolvesToBuilder": func(typeDef ast.Type) bool {
+			_, found := context.ResolveAsBuilder(typeDef)
```
K-Phoen, line 31: There is a `ResolveToBuilder()` function that does exactly what you want :)

internal/jennies/java/converter.go
```diff
@@ -0,0 +1,167 @@
...
+	for _, builder := range context.Builders {
+		output, err := jenny.generateConverter(context, builder)
+		if err != nil {
+			return nil, err
+		}
+
+		filename := filepath.Join(
+			jenny.config.ProjectPath,
+			formatPackageName(builder.Package),
+			fmt.Sprintf("%sConverter.java", tools.UpperCamelCase(builder.Name)),
```
K-Phoen, line 40: Nitpick (doesn't need to get fixed in this PR): we should really use functions like `formatPackageName()`, `formatObjectName()`, ... everywhere

internal/jennies/java/converter.go
```diff
@@ -0,0 +1,167 @@
...
+
+	schema.Objects.Iterate(func(key string, obj ast.Object) {
+		if obj.Type.ImplementedVariant() == string(ast.SchemaVariantDataQuery) {
+			if !obj.Type.HasHint(ast.HintSkipVariantPluginRegistration) {
+				filteredObjects.Set(key, obj)
+			}
+		}
+	})
+
+	return filteredObjects
```
K-Phoen, line 78: How about something like this instead?

```go
return schema.Objects.Filter(func(_ string, obj ast.Object) {
  return obj.IsDataqueryVariant() && !obj.Type.HasHint(ast.HintSkipVariantPluginRegistration)
})
```

internal/jennies/java/converter.go
```diff
@@ -0,0 +1,167 @@
...
+
+		return imports.Add(class, pkg)
+	}
+	typeFormatter := createFormatter(context, jenny.config).withPackageMapper(packageMapper)
+
+	builderNameFormat := fmt.Sprintf("%s.Builder", tools.UpperCamelCase(converter.BuilderName))
+	if isPanel {
+		builderNameFormat = fmt.Sprintf("%s.PanelBuilder", typeFormatter.formatPackage(converter.Package))
+	} else if converter.BuilderName != converter.Input.TypeRef.ReferredType {
+		if converter.Input.TypeRef.ReferredType == variableModel { // TODO: Find a better way to identify this...
```
K-Phoen, line 101: Agreed, let's look at this together: this is going to break with the Dashboard v2 schema :/

also: ideally the jennies should know nothing about the schemas they're generating code for

internal/jennies/java/jennies.go
```diff
@@ -39,6 +40,8 @@ func (config *Config) InterpolateParameters(interpolator func(input string) stri
 func (config Config) MergeWithGlobal(global languages.Config) Config {
 	newConfig := config
 	newConfig.generateBuilders = global.Builders
+	// newConfig.generateConverters = global.Converters
+	newConfig.generateConverters = false
```
K-Phoen, line 44: Just checking: is that left on purpose?

internal/jennies/java/registry.go
```diff
@@ -97,15 +114,22 @@ func (jenny Registry) renderRegistry(context languages.Context) (string, error)
 		return panelSchemas[i].Identifier < panelSchemas[j].Identifier
 	})
 
+	if jenny.config.generateConverters {
+		imports.Add("Panel", "dashboard")
```
K-Phoen, line 118: I'm surprised to see this here: the runtime shouldn't assume that we're generating code related to dashboards/panels

internal/jennies/java/tools.go
```diff
@@ -23,6 +23,14 @@ func formatScalar(val any) any {
 	return newVal
 }
 
+func cleanString(s string) string {
```
K-Phoen, line 26: This function could use a more descriptive name: "clean string" feels pretty vague.

K-Phoen (commented):
This feels almost ready. I'm pretty sure we could get it across the finishing line with a small pair programming session if you're up for it :)

internal/jennies/java/converter.go
spinillos: In the previous comment you said that that function affects to following jennies.

internal/jennies/java/jennies.go
spinillos: Yes, because the converter isn't managing yet unknown dataqueries and panels and "generics" needs more love.

internal/jennies/java/converter.go
K-Phoen: That was because you were reassigning the result of `Filter()` to `schema.Objects`, which is not what I'm suggesting here :)

K-Phoen (approved):
Let's merge this and keep iterating in other PRs, converters are disabled for Java anyway so no risk is taken :)

PR #628, opened 2024-11-05
Allow missing fields in strict unmarshal if they have a default value

K-Phoen:
The strict unmarshal functions in Go are a bit too strict: they return an error if a field is missing from the JSON input, even when that field has a default value.

This PR changes that behavior: required field that have a default value set in the schema can now be omitted.


the-it (approved):
TIL golang templating language 👍

Issue #565, opened 2024-09-27
Missing defaults when are generated in types


spinillos:
Typescript, php and python generate the defaults inside in the "rawtypes" process instead of builders (like go or java). Both solutions work excepts on panels.

Examples of missing defaults: [typescript](https://github.com/grafana/grafana-foundation-sdk/blob/v11.2.x%2Bcog-v0.0.x/typescript/src/candlestick/panelBuilder.gen.ts#L13), [php](https://github.com/grafana/grafana-foundation-sdk/blob/v11.2.x%2Bcog-v0.0.x/php/src/Candlestick/PanelBuilder.php#L15), [python](https://github.com/grafana/grafana-foundation-sdk/blob/v11.2.x%2Bcog-v0.0.x/python/grafana_foundation_sdk/builders/candlestick.py#L18)
Other languages: [go](https://github.com/grafana/grafana-foundation-sdk/blob/v11.2.x%2Bcog-v0.0.x/go/candlestick/panel_builder_gen.go#L918), [java](https://github.com/grafana/grafana-foundation-sdk/blob/v11.2.x%2Bcog-v0.0.x/java/src/main/java/com/grafana/foundation/candlestick/PanelBuilder.java#L39)

You can see that `height` and `span` aren't set in the panels, and the other ones for Options and FieldConfig aren't applied unless to set any of these values.

Issue #527, opened 2024-09-04
Omit an option in veneer, omits the default value for it.


spinillos:
In some languages we generate the default values in the builders and for others, in the types.

When we ignore an option for the builders, and this option contains a default, it's completely skipped. For example, [schemaVersion](https://github.com/grafana/kind-registry/blob/main/grafana/v11.0.x/core/dashboard/dashboard.cue#L78).

For example, in [Go](https://github.com/grafana/grafana-foundation-sdk/blob/v11.0.x%2Bcog-v0.0.x/go/dashboard/dashboard_builder_gen.go#L384) there are less defaults than [Typescript](https://github.com/grafana/grafana-foundation-sdk/blob/v11.0.x%2Bcog-v0.0.x/typescript/src/dashboard/types.gen.ts#L71). In Go we parse the Builders, and Schema in Typescript.

PR #633, opened 2024-11-06
Go/constructors defaults

K-Phoen:
Relates to https://github.com/grafana/cog/issues/565

This PR changes the way default values are dealt with in Go.

Similarly to how these values are handled in PHP, Typescript and Python, defaults are now set in constructors generated for each object.

This change also has the added benefit of ensuring that default values will not be omitted if there is no builder option to set them (see https://github.com/grafana/cog/issues/527).

### DIFF
--- a/internal/jennies/golang/builder.go
+++ b/internal/jennies/golang/builder.go
@@ -10,7 +10,6 @@ import (
 	"github.com/grafana/cog/internal/jennies/common"
 	"github.com/grafana/cog/internal/jennies/template"
 	"github.com/grafana/cog/internal/languages"
-	"github.com/grafana/cog/internal/orderedmap"
 	"github.com/grafana/cog/internal/tools"
 )
 
@@ -69,6 +68,12 @@ func (jenny *Builder) generateBuilder(context languages.Context, builder ast.Bui
 		buildObjectSignature = jenny.typeFormatter.variantInterface(builder.For.Type.ImplementedVariant())
 	}
 
+	constructorName := fmt.Sprintf("New%s", formatObjectName(builder.For.SelfRef.ReferredType))
+	constructorPkg := jenny.typeImportMapper(builder.For.SelfRef.ReferredPkg)
+	if constructorPkg != "" {
+		constructorName = constructorPkg + "." + constructorName
+	}
+
 	jenny.apiRefCollector.BuilderMethod(builder, common.MethodReference{
 		Name: "Build",
 		Comments: []string{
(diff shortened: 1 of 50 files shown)


internal/jennies/golang/builder.go
```diff
@@ -90,105 +95,49 @@ func (jenny *Builder) generateBuilder(context languages.Context, builder ast.Bui
...
-			pkg = opt.Args[i].Type.AsRef().ReferredType
-			_, isBuilder := context.Builders.LocateByObject(opt.Args[i].Type.AsRef().ReferredPkg, pkg)
-			obj, ok := context.LocateObject(opt.Args[i].Type.AsRef().ReferredPkg, pkg)
-			if !ok {
-				return []string{"unknown"}
+	switch typeDef.Kind {
+	case ast.KindRef:
+		resolvedType := context.ResolveRefs(typeDef)
+		if resolvedType.IsStruct() {
+			constructor := fmt.Sprintf("New%s()", typeDef.Ref.ReferredType)
```
jjo, line 123: Nit: ditto using `+` op
```suggestion
			constructor := "New" + typeDef.Ref.ReferredType + "()"
```

internal/jennies/golang/builder.go
```diff
@@ -69,6 +68,12 @@ func (jenny *Builder) generateBuilder(context languages.Context, builder ast.Bui
 		buildObjectSignature = jenny.typeFormatter.variantInterface(builder.For.Type.ImplementedVariant())
 	}
 
+	constructorName := fmt.Sprintf("New%s", formatObjectName(builder.For.SelfRef.ReferredType))
```
jjo, line 71: Nit: using straightfwd `+` for non-formatting string op:

```suggestion
	constructorName := "New" + formatObjectName(builder.For.SelfRef.ReferredType)
```

internal/jennies/golang/builder.go
```diff
@@ -90,105 +95,49 @@ func (jenny *Builder) generateBuilder(context languages.Context, builder ast.Bui
...
-				return []string{"unknown"}
+	switch typeDef.Kind {
+	case ast.KindRef:
+		resolvedType := context.ResolveRefs(typeDef)
+		if resolvedType.IsStruct() {
+			constructor := fmt.Sprintf("New%s()", typeDef.Ref.ReferredType)
+
+			referredPkg := jenny.typeImportMapper(typeDef.Ref.ReferredPkg)
+			if referredPkg != "" {
+				constructor = fmt.Sprintf("%s.%s", referredPkg, constructor)
```
jjo, line 127:
```suggestion
				constructor = referredPkg + "." + constructor
```

internal/jennies/golang/rawtypes.go
```diff
@@ -147,6 +142,154 @@ func (jenny RawTypes) formatObject(schema *ast.Schema, def ast.Object) ([]byte,
 			buffer.WriteString("}\n")
 		}
 	}
+}
+
+func (jenny RawTypes) generateConstructor(buffer *strings.Builder, context languages.Context, object ast.Object) {
+	defName := formatObjectName(object.Name)
```
jjo, line 148: JooC, any reason for not calling this `objName` ?

internal/jennies/golang/rawtypes.go
```diff
@@ -147,6 +142,154 @@ func (jenny RawTypes) formatObject(schema *ast.Schema, def ast.Object) ([]byte,
...
+	}
+
+	if object.Type.IsRef() {
+		referredObj, found := context.LocateObjectByRef(*object.Type.Ref)
+		if !found || !referredObj.Type.IsStruct() {
+			return
+		}
+
+		declareConstructor()
+		buffer.WriteString(fmt.Sprintf("// %[1]s creates a new %[2]s object.\n", constructorName, defName))
```
jjo, line 168: Any particular reason to use `"... %[1]s ... %[2]s ..."` in these strings fmts (here and some places below), instead of just `"... %s ... %s ..."` given that we're using same args order everywhere (and not needed to re-refer to any arg)?

internal/jennies/golang/tmpl.go
```diff
@@ -63,6 +63,7 @@ func initTemplates(extraTemplatesDirectories []string) *template.Template {
 		}),
 		template.Funcs(map[string]any{
 			"formatPackageName": formatPackageName,
+			"formatObjectName":  formatObjectName,
```
jjo, line 66: Mainly for my own understanding: why do we add `formatObjectName` but not `formatFieldName` as template function?
Or maybe I'm failing to "parse" the templates we emit, to find out such difference in tmpl function usage 😇 .

K-Phoen: I started using these `formatX` functions to be consistent with how it's done in other languages than Go: cf https://github.com/grafana/cog/blob/ecbfa726391eb1771f0d08b9bcc057fd6dff07c1/internal/jennies/php/tmpl.go#L25-L34

But since this PR was getting a bit big already, I thought I would change only what I immediately needed and keep working on this in another PR :)

internal/jennies/golang/rawtypes.go
K-Phoen: No real reason, I don't mind changing this if you want :)

internal/jennies/golang/tmpl.go
K-Phoen: The follow-up to this is here: https://github.com/grafana/cog/pull/635 :)

internal/jennies/golang/rawtypes.go
K-Phoen: No good reason: objects used to be called "definitions" a long time ago, and old habits die hard :sweat_smile:

jjo (approved):
LGTM, thanks for the followups!

PR #634, opened 2024-11-06
fix: expose a "data links" option in panel builders

K-Phoen:
Fixes grafana/grafana-foundation-sdk#352

There were two problems causing this issue:

* the option that would allow setting data links was omitted when generating builders (I guess I got confused when I saw two `links` fields on panels :shrug:)
* the `links` field is [defined as `[]any` in the original schema](https://github.com/grafana/kind-registry/blob/2aebf33e52bbfb07a94c44b2359f2b00c630a840/grafana/next/core/dashboard/dashboard.cue#L722)

### DIFF
--- a/config/compiler/kind_registry.yaml
+++ b/config/compiler/kind_registry.yaml
@@ -69,6 +69,16 @@ passes:
             kind: composable_slot
             composable_slot: { variant: dataquery }
 
+  # typed as `links?: [...]` in the original schema
+  - retype_field:
+      field: dashboard.FieldConfig.links
+      as:
+        kind: array
+        array:
+          value_type:
+            kind: ref
+            ref: { referred_pkg: dashboard, referred_type: DashboardLink }
+
   - dashboard_panels: {}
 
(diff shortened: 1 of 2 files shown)


config/compiler/kind_registry.yaml
```diff
@@ -69,6 +69,16 @@ passes:
             kind: composable_slot
             composable_slot: { variant: dataquery }
 
+  # typed as `links?: [...]` in the original schema
```
jjo, line 72: Would there be any needed change here, when upstream original schema gets augmented from current _any_ type for `links`?

jjo (approved):

config/compiler/kind_registry.yaml
K-Phoen: We'd need to remove this change, or it would override the updated upstream schema :)

But since dashboard-related schemas are being revamped, there will likely never be a fix for this. Instead, we'll move over to the "dashboard v2" schemas.

PR #644, opened 2024-11-12
Support fine-grained configuration of Validate()/Equal()/UnmarshalJSO…

K-Phoen:
…NStrict() functions generation in Go

Since cog is going to be used in more contexts than just to generate the Foundation SDK, it now makes sense to be able to configure exactly what should be generated.

Starting with Go and its `Validate()`, `Equal()` and `UnmarshalJSONStrict()` functions on types.

**Question:** how do we want to handle this type of configuration?
Do we prefer having to explicitly enable these features (like this PR does, and like the `GenerateGoMod` option), or explicitly disable them? (see the `SkipRuntime` option)

### DIFF
--- a/config/foundation_sdk.dev.yaml
+++ b/config/foundation_sdk.dev.yaml
@@ -97,6 +97,9 @@ output:
         package_root: '%go_package_root%'
         builder_templates:
           - '%__config_dir%/templates/go'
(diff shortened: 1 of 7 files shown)


K-Phoen:
No changes ↑, that's good!

spinillos (approved):

PR #647, opened 2024-11-13
cue v0.11.0

K-Phoen:
~Bump Cue dependency to v0.11.0-alpha5~
Bump Cue dependency to v0.11.0

We're a bit late to the party when it comes to keeping up to date with Cue, mainly due to a dependency issue that has been fixed in their PR 3475.

~This fix hasn't made it into a non-alpha release, which is why this PR currently depends on `v0.11.0-alpha5` and not `v0.11.0` which hasn't been released yet.~

~I'm planning on waiting until `v0.11.0` is out before merging this. I opened this PR now to keep track of the fixes we need to use that version.~


internal/codegen/cue.go
```diff
@@ -113,18 +113,17 @@ func cueLoader(input CueInput) (ast.Schemas, error) {
...
-		Overlay:    cueFsOverlay,
-		ModuleRoot: "/",
+	bis := load.Instances([]string{"github.com/cog-vfs/" + expectedCuePkgName}, &load.Config{
+		Overlay: cueFsOverlay,
 	})
 
 	value := cuecontext.New().BuildInstance(bis[0])
 	if err := value.Err(); err != nil {
-		return cue.Value{}, err
+		return cue.Value{}, fmt.Errorf("could not building cue instance: %w", err)
```
jjo, line 126:
```suggestion
		return cue.Value{}, fmt.Errorf("could not build cue instance: %w", err)
```

jjo (approved):
LGTM w/small nit

PR #648, opened 2024-11-14
Ignore new dependecy updates for cuelang

spinillos:
We don't want updates from cuelang since everything is fixed to `0.8.0` in Grafana.


K-Phoen:
Hum... we do though :thinking: 

Grafana shouldn't be dictating what version of cue we can or can't use (especially since they way schemas are dealt with is going to change soon in Grafana)

spinillos:
I saw that you created a PR after I create this one 😂. Closing it.

PR #652, opened 2024-11-19
Remove the DashboardPanels compiler pass

K-Phoen:
The same results can be achieved using the existing generic `retype_field` pass.


K-Phoen:
↑ no change in the SDK means that the updated compiler pass config is indeed equivalent to what the custom `DashboardPanels` was doing :)

PR #655, opened 2024-11-20
Remove the `LibraryPanels` compiler pass

K-Phoen:
Similar to #652

The same results can be achieved using generic passes.

<sub>(yes, my goal is to get rid of all these overly specific passes... or as many of them as I can :sweat_smile:)</sub>


K-Phoen:
The big diff was expected: I renamed the `LibrarypanelLibraryPanelModel` to `PanelModel`, which means some files also got renamed

spinillos (approved):

PR #661, opened 2024-11-22
Support customizing how enums should be generated in Typescript

K-Phoen:
A `EnumsAsUnionTypes` configuration option is introduced, to generate enums as a union of values instead of using an actual `enum` declaration.

If it is false, an enum will be generated as:

```ts
enum Direction {
  Up = "up",
  Down = "down",
  Left = "left",
  Right = "right",
}
```

If it is true, the same enum will be generated as:

```ts
type Direction = "up" | "down" | "left" | "right";
```

This option will be useful in the context of the dashboard V2 schema in grafana/grafana, where enums as disjunctions of concrete values are preferable to actual `enum` declarations.

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -186,6 +186,23 @@ func (pipeline *SchemaToTypesPipeline) Golang(config GoConfig) *SchemaToTypesPip
 type TypescriptConfig struct {
 	// ImportsMap associates package names to their import path.
 	ImportsMap map[string]string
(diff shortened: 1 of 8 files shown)


K-Phoen:
↑ no change expected to the Foundation SDK since this new option isn't used by default.

spinillos (approved):

Issue #653, opened 2024-11-19
Experiment with the Dashboard V2 schema


K-Phoen:
Can be found in grafana/grafana: https://github.com/grafana/grafana/blob/main/packages/grafana-schema/src/schema/dashboard/v2alpha0/dashboard.schema.cue

Validate that:

* [ ] cog can parse it
* [ ] cog generates valid code from it (types, builders, converters)
  * [ ] in TS
  * [ ] in Go
  * [ ] in PHP
  * [ ] in Java
* [ ]  we can use compiler passes/veneers to generate good builders from it

PR #667, opened 2024-11-26
Make the ComposeDashboardPanel builder veneer configurable

K-Phoen:
Relates to #653

In order to properly support the Dashboard v2 schema in the Foundation SDK, we need to be able to configure how to compose the various "panel plugins" into self-contained panel builders.

This PR takes the builder veneer that we use today and makes it configurable enough to be used for Dashboard v2 as well.


K-Phoen:
↑ no change means I haven't completely borked panels composition :sweat_smile:

spinillos (approved):

PR #669, opened 2024-11-28
Hidden constant variables

K-Phoen:
Relates to grafana/grafana-foundation-sdk#391

Took a bit more effort than it should since enum values were not always correctly formatted in constructor assignments.

### DIFF
--- a/internal/ast/types.go
+++ b/internal/ast/types.go
@@ -672,6 +672,10 @@ func (t EnumType) MemberForValue(value any) (EnumValue, bool) {
 		return EnumValue{}, false
 	}
 
+	if value == nil {
+		return t.Values[0], false
+	}
+
 	equal := func(a, b any) bool {
 		return a == b
 	}
--- a/internal/ast/types.go
+++ b/internal/ast/types.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/grafana/cog/internal/tools"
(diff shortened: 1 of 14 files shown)


internal/ast/types.go
```diff
@@ -666,6 +667,33 @@ type EnumType struct {
 	Values []EnumValue // possible values. Value types might be different
 }
 
+func (t EnumType) MemberForValue(value any) (EnumValue, bool) {
+	if len(t.Values) == 0 {
+		return EnumValue{}, false
+	}
+
+	if value == nil {
+		return t.Values[0], false
```
spinillos, line 676: Its assuming that its the default one?

internal/ast/types.go
```diff
@@ -666,6 +667,33 @@ type EnumType struct {
...
+	}
+
+	equal := func(a, b any) bool {
```
spinillos, line 679: What is the benefit of that function?

K-Phoen: To work around a kink caused by using `any` in our IR. Because of that, some numerical values are stored as an `int` (within an `any` field) while others can be `int64`, `int32`, ...
Which means that for now, in order to have equality checks that work as expected, we need to do some weird things :/

internal/ast/types.go
K-Phoen: Kinda but not really, it's just to return "something" that can be used as a valid empty value for an enum.

PR #680, opened 2024-12-11
Make the ComposeDashboardPanel builder veneer fully generic

K-Phoen:
This veneer was usable only for dashboard panels. With the dashboard v2 schema, it is very likely that we'll need to use it for queries too.

This PR turns the `ComposeDashboardPanel` veneer into a fully generic `Compose` one.


K-Phoen:
The failure to generate the preview was expected: the veneer config changed between this branch and main :shrug:

spinillos (approved):

PR #683, opened 2024-12-11
Maps of builders

K-Phoen:
Given the following schema:

```cue
Panel: {
  title: string
}

Dashboard: {
  panels: {
    [string]: Panel
  }
}
```

One would expect a DashboardBuilder to be generated with the following option:

```go
func (builder *DashboardBuilder) Panels(panels map[string]cog.Builder[Panel]) {
  // …
}
```

Without this PR, the following option is generated:

```go
func (builder *DashboardBuilder) Panels(panels map[string]Panel) { // note: map of panels, not panel builders
  // …
}
```

ToDo:

* [x] ~java~ → https://github.com/grafana/cog/issues/684
* [x] update go converters
* [x] update PHP converters

### DIFF
--- a/internal/jennies/golang/converter.go
+++ b/internal/jennies/golang/converter.go
@@ -83,11 +83,10 @@ func (jenny *Converter) generateConverter(context languages.Context, builder ast
 			"importStdPkg": func(pkg string) string {
 				return imports.Add(pkg, pkg)
 			},
-			"importPkg":           typeImportMapper,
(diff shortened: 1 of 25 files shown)


K-Phoen:
The diff is somewhat expected.
To avoid breaking BC in the Foundation SDK, we can define a veneer to not generate a builder for the `ValueMappingResult` object (which doesn't need one anyway as it's a fairly small type)

PR #691, opened 2024-12-16
Github: Experimenting with changesets

spinillos:
Changesets "allows" to use other language than javascript: [doc](https://github.com/changesets/changesets/blob/main/docs/versioning-apps.md).

The requirement is to create a minimum `package.json` configuration file and we create our own CI flow.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,7 +2,10 @@ name: Release
 
 on:
   push:
-    tags: ['v*']
+    branches:
+      - main
+        
+concurrency: ${{ github.workflow }}-${{ github.ref }}
 
 permissions:
   contents: write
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -11,9 +14,23 @@ env:
...
+        
+      - name: Create Release Pull Request
+        uses: changesets/action@c8bada60c408975afd1a20b3db81d6eee6789308 #v1.4.9
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+    
   release:
     name: Release
     runs-on: ubuntu-latest
+    if: steps.changesets.outputs.published == 'true'
```
K-Phoen, line 33: That doesn't look very valid :thinking:

K-Phoen:
This has been inactive for a while: closing.

PR #712, opened 2025-01-16
Add an option allowing to emit `any` types as `interface{}` in Go

K-Phoen:
Will be useful in codebases that don't support `any` (or prefer `interface{}`).


IfSentient (approved):
Thanks for putting this together so quickly!

PR #714, opened 2025-01-20
Bootstrap the docs

K-Phoen:
This PR introduces some documentation on codegen pipelines, accepted input schemas, outputs, transformations, ...

Everything isn't documented yet, the structure of the docs could definitely be better, ... but the idea behind this PR is to introduce *some* documentation. A foundation to improve and iterate on, as opposed to having nothing and postponing writing docs for anything because the cost of getting started is too great :sweat_smile: <sub>which is definitely not the reason why we didn't have any docs until this PR. Not at all :trollface: </sub>


jjo:
> This PR introduces some documentation on codegen pipelines, accepted input schemas, outputs, transformations, ...
> 
> Everything isn't documented yet, the structure of the docs could definitely be better, ... but the idea behind this PR is to introduce _some_ documentation. A foundation to improve and iterate on, as opposed to having nothing and postponing writing docs for anything because the cost of getting started is too great 😅 which is definitely not the reason why we didn't have any docs until this PR. Not at all :trollface:

Thanks for this!
During our meeting today we discussed some bits to possibly improve the main `index.md` file, although they not seem to be included here, maybe keep them as TODO-s inside `<!-- ... -->` MD comment blocks?

K-Phoen:
I just skipped the comparison with protoc for now: I'm not sure how to word it yet and I wanted to get this PR moving before it hit 2k lines added :sweat_smile:

jjo:
> I just skipped the comparison with protoc for now: I'm not sure how to word it yet and I wanted to get this PR moving before it hit 2k lines added 😅

+1 apologies somehow I overlooked e.g. the `use-cases` section you added, LGTM

jjo (approved):

PR #664, opened 2024-11-25
Sandbox/dashboard v2

K-Phoen:
Relates to #653

:warning: **Not meant to be merged** :warning: 

Just a playground to experiment and get more familiar with the Dashboard v2 schema.

#### Useful commands

Generating the code (for dashboards v2):

```shell
rm -rf generated && devbox run go run cmd/cli/main.go generate \
        --config ./configv2/foundation_sdk.dev.yaml
```

Running the Typescript example:

```shell
devbox run ts-node ./examples/typescript_dashboardv2/index.ts
```

Running the Go example:

```shell
devbox run go run ./examples/_go_dashboard_v2/*.go
```

PR #722, opened 2025-01-24
Sandbox/dashboard v1 v2

K-Phoen:
Relates to #653

Spiritual successor of #664


internal/jennies/php/jennies.go
```diff
@@ -158,6 +158,8 @@ func (language *Language) CompilerPasses() compiler.Passes {
 		&compiler.FlattenDisjunctions{},
 		&compiler.DisjunctionInferMapping{},
 		&compiler.UndiscriminatedDisjunctionToAny{},
+		&compiler.DisjunctionToType{},
+		&compiler.RemoveIntersections{},
```
K-Phoen, line 162: This is making my eyebrows go up :laughing: 

Why are these compiler passes needed? I'm especially worried about the `DisjunctionToType` one: it shouldn't be needed and using it will actually introduce a bunch of completely avoidable BC breaks in the foundation SDK

spinillos: I added because it "fixes" the converter in PHP.  One of the affected schemas is Expr, but I think that is because this [veneer](https://github.com/grafana/grafana-foundation-sdk/blob/main/.cog/compiler/expr_passes.yaml). 

The problem is that without `DisjunctionToType`, the disjunction appears in the `guard` block ([here](https://github.com/grafana/cog/blob/main/internal/jennies/php/templates/converters/converter.tmpl#L17)), instead of the BuilderDisjunction ([here](https://github.com/grafana/cog/blob/main/internal/jennies/php/templates/converters/converter.tmpl#L64)), making it more complex.  

The second compiler pass is to remove aliases (like Java).

Maybe we have any bug in the veneer? It also applies `disjunction_to_type`, but at the beginning.

Issue #715, opened 2025-01-20
Set enum values into a field, uses enum types and enum isn't created.


spinillos:
When we have enums defined in cue and used as `Enum & "value"`, the generated code set the fields as the type of the enum skipping its generation.

Its happening with this [enum](https://github.com/grafana/grafana/blob/main/kinds/dashboard/dashboard_kind.cue#L379).

PR #735, opened 2025-01-28
Default value enum and types

spinillos:
Closes https://github.com/grafana/cog/issues/715

We weren't identifying values that set a specific enum value as enums. Instead, we set these fields as integers or strings directly.

It adds a new check in cue to know if we need to reference to the specific enum and "force" the default as the defined value. Because is isn't set with the asterisk (to identify a default in cue), we should parse the value as string or integer directly and set as default.

<details>
  <summary>If you need cue a bit more explained👇🏼 </summary>
  
  When we have an enum like:
  ```cue
  myEnum: "a" | "b" | "c"
  ```
  and we have a field set as:
  ```cue
  field: myEnum & "a"
  ```
  It means that we want to use the value "a" of the enum. The problem is that `field` was set as `string`, instead of `MyEnum.A`

  For cue, the value "a" isn't a default since a default is defined with the asterisk like this:
  ```cue
  field: *"a" | "b" | "c"
  ```
  And for the enum case, we need to say that "a" is the default that we want to force to use it.
</details>

The side effect of this, is that we have some disjunctions with values that points into this new enums and we weren't able to infer the discriminator and the mapping because they aren't scalars. So it needs to update the veneer to identify them.


spinillos:
Closing it since it isn't the way to do it.

PR #745, opened 2025-02-03
Better detection of "out-of-root" types in CUE

K-Phoen:
Relates to grafana/grafana#99966

One issue highlighted by the PR above comes from cog not being always able to discover objects within a schema (especially objects defined outside of the cue value given as input to cog).

To fix that, the CUE parser now follows references and parses the objects they point to.

### DIFF
--- a/internal/simplecue/generator.go
+++ b/internal/simplecue/generator.go
@@ -113,12 +113,9 @@ func (g *generator) walkCueSchemaWithEnvelope(envelopeName string, v cue.Value) 
 	for i.Next() {
 		if i.Selector().IsDefinition() {
 			name := g.namingFunc(g.rootVal, i.Value().Path())
-			n, err := g.declareObject(name, i.Value())
-			if err != nil {
+			if err := g.declareObject(name, i.Value()); err != nil {
 				return err
 			}
-
-			g.schema.AddObject(n)
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


K-Phoen:
The huge diff in https://github.com/grafana/cog/pull/745#issuecomment-2631909017 is because objects in a schema are discovered in a different order with this PR :|

internal/simplecue/generator.go
```diff
@@ -490,6 +476,14 @@ func (g *generator) declareReference(v cue.Value, defV cue.Value) (ast.Type, err
 			})), nil
 		}
 
+		// ensure that referenced objects are explored (in case they're not
+		// defined within the "root" cue value given to cog as parsing input)
+		if refPkg == g.schema.Package && !g.schema.Objects.Has(refType) {
+			if err := g.declareObject(refType, referenceRootValue.LookupPath(path)); err != nil {
```
K-Phoen, line 482: This is the actual fix :)

internal/simplecue/generator.go
```diff
@@ -113,12 +113,9 @@ func (g *generator) walkCueSchemaWithEnvelope(envelopeName string, v cue.Value)
 	for i.Next() {
 		if i.Selector().IsDefinition() {
 			name := g.namingFunc(g.rootVal, i.Value().Path())
-			n, err := g.declareObject(name, i.Value())
-			if err != nil {
+			if err := g.declareObject(name, i.Value()); err != nil {
 				return err
 			}
-
-			g.schema.AddObject(n)
```
Pokom, line 121: I don't `g.schema.AddObject(n)` called anywhere else. Was this removed intentionally?

Pokom (approved):
Just one question about the refactored parts, but fix looks 👍🏻

internal/simplecue/generator.go
K-Phoen: It's done directly in `declareObject()` now.

I found it a bit misleading to have a "declareObject" function that didn't actually add the object to the schema :sweat_smile:

Pokom: Ah okay! I did see that called, but wasn't sure if it was the equivalent. Agreed this is much clearer to have it done once, then in each iteration.

Issue #778, opened 2025-03-11
Allow lenient marshaling of disjunctions


radiohead:
### What

We'd like to have `cog` codegen support lenient marshaling of disjunctions in Go. Currently generated JSON marshal / unmarshal code will throw errors if none of the values were set.

We can (and should) make it opt-in by providing a config option to enable it, while keeping the default intact.

### Why

Currently Go code generated by e.g. https://github.com/grafana/grafana/pull/99966 results in panics, due to the specifics of how Grafana API server and OpenAPI generation is set up.

Having the ability to bypass those disjunction errors lets us sidestep this issue until we figure out how to properly enforce schema validation.

radiohead:
I've discussed this with @K-Phoen – I'm working on a PR to address this.

K-Phoen:
Let me know if you need anything :)

PR #779, opened 2025-03-13
Add an option to allow marshaling empty Go disjunctions

radiohead:
### What

This change adds a new config option for Go JSON marshaling Jenny, which makes generated code ignore empty disjunctions when marshaling.

### Why

There is a use-case for it in `grafana/grafana` where returning an error from the marshaler causes the API to panic, due to how schema validation is enforced on the server.

Closes https://github.com/grafana/cog/issues/778


K-Phoen:
I tweaked the tests a bit to not need to add thousands of lines of additional test files.

The test I added instead isn't perfect, but we're kinda reaching the limits of our test suite for jennies when it comes to dealing with the options we keep adding. We can revisit that setup later though.

K-Phoen (approved):

PR #785, opened 2025-03-28
Support disabling post-formatting in Go jenny

K-Phoen:
If the generated Go code isn't valid, post-formatting will fail and prevent the code from being written to disk.

To make debugging these cases easier, this PR adds an option to disable this formatting.


K-Phoen:
The tests are failing because of a broken schema for the `next` version of Grafana

spinillos (approved):

PR #798, opened 2025-04-21
Add OpenAPI output to the SchemaToTypesPipeline

IfSentient:
This PR adds the ability to output OpenAPI JSON from the `SchemasToTypesPipline` type, using the `GenerateOpenAPI` method. Method is named `GenerateOpenAPI` rather than `OpenAPI` (like the go and TypeScript variants), as I wanted to make sure it was distinct from a possible future input type of OpenAPI (like the current `CUEValue` method that sets the input). 

OpenAPI output can be used by the [grafana-app-sdk](https://github.com/grafana/grafana-app-sdk) instead of CUE's openAPI encoder to allow the app-sdk to parse recursive types from CUE (which CUE's openAPI encoder fails on).

### DIFF
--- a/helpers.go
+++ b/helpers.go
@@ -6,7 +6,6 @@ import (
 
 	"cuelang.org/go/cue"
 	"github.com/grafana/codejen"
-
 	"github.com/grafana/cog/internal/ast/compiler"
 	"github.com/grafana/cog/internal/codegen"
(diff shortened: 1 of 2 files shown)


K-Phoen (approved):
LGTM :+1: 

Glad that this openapi output can be useful in grafana-app-sdk :)

PR #808, opened 2025-05-19
Go: Fix error propagation

spinillos:
Fixes https://github.com/grafana/grafana-foundation-sdk/issues/748

It returns builder.errors instead of nil when validation is ok. The error happens in panels because they extend from the Panel and this validator doesn't check internal values.

The error map was changed into an array since the index value is never used.

### DIFF
--- a/examples/_go_alerting/main.go
+++ b/examples/_go_alerting/main.go
@@ -21,6 +21,7 @@ func main() {
 		Interval(alerting.Duration(interval.Seconds())). // type is confusing
 		WithRule(alerting.NewRuleBuilder("cog alert rule name").
 			For("5m").
+			RuleGroup("group").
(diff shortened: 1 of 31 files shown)


Duologic (approved):

K-Phoen:
I haven't read this PR in details but it doesn't look like it's addressing the root of the issue.

From what I see in [`Panel.Validate()`](https://github.com/grafana/grafana-foundation-sdk/blob/next%2Bcog-v0.0.x/go/dashboard/types_gen.go#L1787), we never call the validation function for panels options/fieldConfig structs.

This bug makes sense since these fields are defined as "any" in the schemas, so cog doesn't know that they are always populated with structs coming from panel plugins and calling `Validate()` is safe.

PR #809, opened 2025-05-29
Pipeline: Create output options with replace extensions map

spinillos:
Its a dummy change that allows to update the extension of the parsed files for CI configuration. This configuration is specific per language that adds information for testing and publishing.

For that we are using templates where we replace some information to set the proper version of Grafana since we are working in branches and each branch, should have its version. The problem is that we are using `yaml` files as templates and these templates have escape values to be able to set this information and its detected as "invalid" with `zizmor`.

So with this change we can "skip" this check using a different format against the templates. When we create a new release, the template will generate the final yaml file for CI and `zizmor` will be executed.

### DIFF
--- a/internal/codegen/output.go
+++ b/internal/codegen/output.go
@@ -41,6 +41,9 @@ type Output struct {
 	// TemplatesData holds data that will be injected into package and
 	// repository templates when rendering them.
 	TemplatesData map[string]string `yaml:"templates_data"`
+
+	// OutputOptions configures the output of the file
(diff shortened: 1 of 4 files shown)


spinillos:
@K-Phoen there isn't a way to ignore complete files. You can add some rules to skip some validations, but it still fails because we are using the yaml files in foundaiton-sdk as templates and it detects them as "invalid" since we added some escape functions to make it works with Go. Its why the reason of this change...

Duologic (approved):

PR #814, opened 2025-06-13
Go: Add YAML struct tags to generated code

### LINKED ISSUE #812: Go: Add `yaml` struct tags
Cog generates types with `json` struct tags:
```go
type GeneratedStruct struct {
  FieldOne int `json:"fieldOne"`
}
```

Several popular Go yaml libraries need `yaml` tags set as well to serialize and deserialize correctly. It would be nice to add those to the generated code.

joeblubaugh:
This supports users who want to deserialize to Go using common YAML libraries like gopkg.in/yaml.v3

Fixes #812

### DIFF
--- a/internal/jennies/golang/builder_test.go
+++ b/internal/jennies/golang/builder_test.go
@@ -99,7 +99,7 @@ func TestBuilder_emptyValueForGuard(t *testing.T) {
 			context: languages.Context{},
 			input:   ast.NewStruct(ast.NewStructField("field", ast.String())),
 			expected: `&struct {
-    Field string ` + "`" + `json:"field,omitempty"` + "`" + `
+		Field string ` + "`" + `json:"field,omitempty" yaml:"field,omitempty"` + "`" + `
 }{}`,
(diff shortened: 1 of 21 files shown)


spinillos:
If this change is only for app-platform needs, I suggest you to add a flag in the client [here](https://github.com/grafana/cog/blob/main/helpers.go#L183) and pass it to the jenny and set if its enabled. This change affects to whole `foundation-sdk`.

If you execute `rawtypes_test.go` to know if everything is ok is enough because it only adds yaml tag.

joeblubaugh:
Good point about the custom unmarshallers - that complicates things enough that this PR should probably be closed.

I think support for YAML as a serialization is probably very useful for many people. An option might be to "officially" support sigs.k8s.io/yaml, which is a maintained fork of the unmaintained ghodss/yaml and also has the behavior you note of using `json` tags for naming fields, and gets the custom unmarshalling behavior. That's a no-op for this repository, except maybe some optional documentation of that fact.

PR #821, opened 2025-06-24
Java: Custom deserializer fixes

spinillos:
Java didn't support custom Deserializers (ones added by config). This is necessary to make it works with V2 schemas since we have to add custom schemas for some models.

Also it removes the builder check in factory constructors. It doesn't have sense because types shouldn't know about builders.

Note: There are some code that still living in cog that should be move into `foundation-sdk` since is specific to schemas.


the-it (approved):
unblocking ... lgtm

PR #826, opened 2025-07-07
[CUE->OpenAPI] Add Manual Open Attribute for CUE -> OpenAPI Generation

IfSentient:
Add hints to AST struct type, and use the 'open' hint to add `additionalProperties` in openAPI generation.

This allows an author of a CUE schema to specify if a type should be treated as open in scenarios where it is not simply
```cue
foo: {
    [string]: any
}
```
so now something like
```
foo: {
    bar: string
    [string]: any
} @cog(open=true)
```
will generate openAPI which has the `bar` required property, and `additionalProperties: {}` to allow for additional fields.

### DIFF
--- a/internal/jennies/jsonschema/schema.go
+++ b/internal/jennies/jsonschema/schema.go
@@ -202,6 +202,11 @@ func (jenny Schema) formatStruct(typeDef ast.Type) Definition {
 
 	definition.Set("type", "object")
 	definition.Set("additionalProperties", false)
+	if typeDef.HasHint("open") {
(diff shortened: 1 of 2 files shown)


spinillos:
You must to get the annotation in the cue parser like [this](https://github.com/grafana/cog/blob/main/internal/simplecue/generator.go#L242) and define the hint [here](https://github.com/grafana/cog/blob/main/internal/ast/hints.go).

spinillos (approved):

PR #829, opened 2025-07-16
Passes: Add enum value action

spinillos:
It adds an action to be able to add more fields into an enum by configuration. It applies to explicit defined enum (like a normal definition) and for enums declared directly into a field.

Related issue: https://github.com/grafana/grafana-foundation-sdk/issues/812


K-Phoen:
Closing this since the issue could be solved using existing rewrite rules.

PR #840, opened 2025-08-27
Go: Add missing unmarshal templates for multiple types

spinillos:
It adds templates for disjunctions of multiple kinds. Without them we were using the default unmarshal generating producing invalid code when we unmarshal a Json.

I also fixed a NPE when we try to prefix a enum without set a value.


joeblubaugh:
I think this PR would be stronger with some tests of the various cases. Maybe I'm not understanding how the testdata/ files work correctly, but I don't see cases of disjunctions of multiple non-scalar types in there.

spinillos:
@joeblubaugh could you test if this solution works for you? Its a bit annoying manage disjunctions with multiple kind of fields. In the worst case I can undo the change that adds these unmarshal functions for those kind of disjunctions.

spinillos:
I usually test it adding a replace rule in go.mod pointing to my local cog repository in app-sdk, and install the library.

PR #853, opened 2025-09-18
Fix Constant References null checks

spinillos:
Edited!!!

* **Go**: It fixes the fields declared as ConstantReferences when they are optional. We were missing to set them as nullable, making the equal function to fail.

* **Typescript**: We weren't setting the ConstantReference value in the constructor when it was optional.

* **All builders**: We were skipping the options declared as ConstantReference and they were used as fixed value. The option now is available only if the field is optional.


radiohead (approved):

spinillos:
@radiohead You are right... the constant ref isn't a discriminator here so my change isn't correct.

PR #1070, opened 2026-04-10
Fix breaking changes from cue upgrade to 0.16

moustafab:
This fixes issues which arise from the upgrade to cue 0.16 from 0.11. The big issue surrounded handling arrays (lists) with defaults being mishandled by our AST parsing due to a change to how `value.Allows` works since v0.15 [(relevant commit)](https://github.com/cue-lang/cue/commit/a6f97a13c8314d7c2ab232bbfb967c7f7bc18160)

The solution is outlined in the three commits, but essentially we peel off the default and make sure we detect the openness of a list more thoroughly.


moustafab:
Afaict the diff output for the foundation sdk stuff is a no-op, just re-ordering but someone with more familiarity should confirm.

PR #1079, opened 2026-04-16
Cli: Disjunctions without discriminator option

spinillos:
It creates an option to generate disjunctions without discriminator as the same way as we have it when we have a mix of types in a disjunction.

There are some Grafana apps contains disjunction without discriminator (because at the moment they don't want to change the current model) and they want to generate the proper Marshal and Unmarshal functions. For that it uses `json.NewDecoder` that tries to match the object to its proper object.

It's under a flag in the client because we don't want to change the behaviour of Foundation-SDK generation (at the moment).


internal/ast/compiler/undiscriminated_disjunctions_to_type.go
```diff
@@ -0,0 +1,93 @@
...
+		}
+
+		processedBranch := branch
+		processedBranch.Nullable = true
+
+		fields = append(fields, ast.NewStructField(ast.TypeName(processedBranch), processedBranch))
+	}
+
+	structType := ast.NewStruct(fields...)
+	structType.Hints[ast.HintUndiscriminatedDisjunctionOfRefs] = disjunction
```
radiohead: Looks like this is very similar to `DisjunctionToType.processDisjunction`, could we refactor & extract?

This is also missing `maps.Copy(structType.Hints, def.Hints)` (present in `DisjunctionToType`) – I assume by mistake?

internal/ast/hints.go
```diff
@@ -36,6 +36,11 @@ const (
 	// when generating outputs which make explicit allowances for
 	// arbitrary additional properties, such as OpenAPI/JSONSchema.
 	HintOpenStruct = "open"
+
+	// HintUndiscriminatedDisjunctionOfRefs indicates that the struct was
+	// previously represented in the IR by a disjunction of references to structs
+	// with no discriminator field.
+	HintUndiscriminatedDisjunctionOfRefs = "undiscriminated_disjunction_of_refs"
```
radiohead, line 44: IIUC this is Go-only currently? Should we leave a comment about that or whatever?

radiohead (commented):
Overall LGTM, only one concern about missing hints copy in the new compiler pass.

internal/ast/hints.go
spinillos: I only added in Go because we want to use it for the client (at the moment). But if we want to apply for Foundation-SDK, Java needs it too.

radiohead: I'm OK with keeping it Go-only for now but I'd add a comment about it.

internal/ast/compiler/undiscriminated_disjunctions_to_type.go
spinillos: I refactored and merged it into `DisjunctionToType`.

IfSentient (approved):

PR #1081, opened 2026-04-17
Cue: Enums as map index

### LINKED ISSUE #893: CUE: Use enums as map index, produces generic output
Having:
```cue
MyEnum: "a" | "b" | "c"

MyStruct: {
  val: map[MyEnum]bool
}
```

produces (Go):
```go
Val interface{} `json"val"`
```

instead of:
```go
Val map[MyEnum]bool `json:"val"`
```

This error is in all languages.

spinillos:
Closes: https://github.com/grafana/cog/issues/893

* It allows cue parser to identify enums as map index to generate them properly in all languages.
* Update hardcoded String in Java maps.
* Skip factory generation in Typescript if we have maps with this kind of index and without defaults.


internal/simplecue/utils.go
```diff
@@ -268,6 +268,46 @@ func isImplicitEnum(v cue.Value) (bool, error) {
 	return true, nil
 }
 
+// bulkOptionalFieldFromSyntax finds the first bulk-optional field [X]: T in v.
+// Returns the pattern expression X and the value expression T.
+//
+// It first tries v.Syntax() (the CUE Def export), but falls back to v.Source()
+// when Syntax() returns a *ast.BadExpr — which happens for enum-indexed maps
+// due to a CUE exporter bug: "reference in label expression refers to field
+// against which it would be matched" (see https://cuelang.org/issues/new).
```
IfSentient: The comment just refers to the open a new issue page in the cue repo, I think it should be referring to https://github.com/cue-lang/cue/issues/4291, though I may have found the wrong issue searching.

IfSentient (commented):
LGTM, just one note about the CUE issue in the go comments

PR #1083, opened 2026-04-21
Java: Update composable class names

spinillos:
I don't recall the reason why, but Java was using the original class name for composable. This was generating conflicts when we try to add dashboard v2 schema because they are using the same fields and same names for composable, generating name conflicts.

With this change, it will use the name defined in the composable, following the same pattern as the rest of languages and fixing the issue.

We need it to make https://github.com/grafana/grafana-foundation-sdk/pull/1191 to work.


radiohead (approved):

harisrozajac (approved):
🙌

Issue #837, opened 2025-08-25
Features consistency


spinillos:
We usually add new features to Go, after that to Typescript, and eventually to the rest of the languages. We add them in this order because the needs that we have internally in Grafana.

They can be done in any order, except Typescript converter since it misses the registry.

- Equal functions ([Reference](https://github.com/grafana/cog/blob/main/internal/jennies/golang/equality.go)):
  - [ ] Java
  - [ ] PHP
  - [ ] Python
  - [ ] Typescript
- Converters ([Reference](https://github.com/grafana/cog/blob/main/internal/jennies/golang/converter.go))
  - [ ] [Java](https://github.com/grafana/cog/issues/591) (in progress)
  - [ ] [Python](https://github.com/grafana/cog/issues/590)
  - [ ] [Typescript](https://github.com/grafana/cog/issues/592)
- Registry (necessary for converters. [Reference](https://github.com/grafana/grafana-foundation-sdk/blob/main/.cog/templates/go/extra/cog/plugins/variants.go))
  - [ ] [Typescript](https://github.com/grafana/cog/issues/593) 
- Validation functions ([Reference](https://github.com/grafana/cog/blob/main/internal/jennies/golang/validation.go))
  - [ ] [Java](https://github.com/grafana/cog/issues/772)
  - [ ] [PHP](https://github.com/grafana/cog/issues/774)
  - [ ] [Python](https://github.com/grafana/cog/issues/773)
  - [ ] [Typescript](https://github.com/grafana/cog/issues/775)
- Strict Json ([Reference](https://github.com/grafana/cog/blob/main/internal/jennies/golang/strictjson.go))
  - [ ] Java
  - [ ] PHP
  - [ ] Python
  - [ ] Typescript

PR #1084, opened 2026-04-22
Equal functions

spinillos:
It adds equal function for Java, PHP, Python and Typescript to have the same equal feature than Golang.

It contributes to: https://github.com/grafana/cog/issues/837

There isn't any diff because equal option have to be setup in Foundation-SDK config...


internal/jennies/typescript/equality.go
```diff
@@ -0,0 +1,99 @@
...
+func (jenny equalityMethods) generateForObject(context languages.Context, object ast.Object) string {
+	if !object.Type.IsStruct() {
+		return ""
+	}
+
+	objectName := formatObjectName(object.Name)
+	var buffer strings.Builder
+
+	buffer.WriteString(fmt.Sprintf("// equals%s tests the equality of two `%s` objects.\n", objectName, objectName))
+	buffer.WriteString(fmt.Sprintf("export const equals%s = (a: %s, b: %s): boolean => {\n", objectName, objectName, objectName))
```
IfSentient: Would it make sense to use `jenny.equalsFuncName` here to ensure consistency in naming?

IfSentient (approved):
LGTM, just one nit about using the function for ensuring consistent naming in the TypeScript codegen.

PR #1091, opened 2026-04-28
chore(deps): Upgrade kin-openapi to v0.137.0

IfSentient:
Upgrade `getkin/kin-openapi` to `v0.137.0`. This update included a breaking change around handling `exclusiveMinimum` and `exclusiveMaximum`, as [they are handled differently in OpenAPI 3.0 vs 3.1](https://www.openapis.org/blog/2021/02/16/migrating-from-openapi-3-0-to-3-1-0). 

Added tests to check for both 3.0 and 3.1 behavior compliance.


internal/openapi/utils.go
```diff
@@ -59,25 +59,39 @@ func getConstraints(schema *openapi3.Schema) []ast.TypeConstraint {
...
 			Op:   op,
-			Args: getArgs(schema.Min, schema.Type.Slice()[0]),
+			Args: getArgs(min, schema.Type.Slice()[0]),
 		})
 	}
 
-	if schema.Max != nil {
+	if schema.Max != nil || schema.ExclusiveMax.Value != nil {
+		// In openapi3.0, schema.Max is used alongside a boolean exclusiveMaximum.
+		// In 3.1, exclusiveMaximum is the actual value, rather than the boolean modifying maximum
```
spinillos, line 82: Are we supporting 3.1 in app-sdk? Mainly because we only generate the schema for [3.0.0](https://github.com/grafana/cog/blob/main/internal/jennies/openapi/schema.go#L61)

IfSentient: We don't support codegen from any OpenAPI types in app-sdk at the moment, only CUE. I just figured I might as well have the fix be forwards-compatible in the cog side, especially since it would just be silently ignored if we drop the constraint when `schema.ExclusiveMax.IsTrue()` is false.

spinillos (approved):

PR #1098, opened 2026-05-01
feat(csharp): add C# code generation target

fdrstrok:
## Summary

Adds **C#** as a new code-generation target to `cog`, sitting alongside Go, Java, PHP, Python, TypeScript, Terraform, JSON Schema and OpenAPI. The implementation closely mirrors the Java jenny — the closest existing OO/strongly-typed analogue — and produces idiomatic .NET 10 C# code with nullable reference types enabled.

This is the foundational work to enable a future C# Grafana Foundation SDK.

## What's generated

For each schema, the new jenny emits:

- **Raw types** — `class`, `enum` and constant declarations under `src/<NamespaceRoot>/<Package>/<Type>.cs`. Default namespace root is `Grafana.Foundation`, configurable via `namespace_root`.
- **`System.Text.Json` converters** (opt-in via `generate_json_converters: true`) — per-field `[JsonPropertyName]`, per-class `[JsonConverter(typeof(<T>JsonConverter))]`, per-string-enum `[JsonConverter(typeof(JsonStringEnumConverter<T>))]` + `[JsonStringEnumMemberName(...)]`. Disjunction shapes (refs/scalars/mixed) get bespoke `JsonConverter<T>` classes mirroring Java's `couldBe` semantics.
- **Fluent builders** (gated on the global `output.builders` flag) — `public class <Name>Builder : Cog.IBuilder<TargetType>` with chainable PascalCase option methods, constructor + property fields, nil-checks for nested parents, builder-typed args via `Cog.IBuilder<T>` / `List<Cog.IBuilder<T>>` / `Dictionary<string, Cog.IBuilder<T>>` unfolded into plain collections via `foreach + .Build()`. Constraints (`>=`, `<=`, regex) throw `System.ArgumentException`.
- **Cog runtime** — single `Cog/IBuilder.cs` with `public interface IBuilder<out T> { T Build(); }`. Skippable via `skip_runtime: true` (mutually exclusive with builders).

Composable slots render as `Cog.Variants.<Variant>` (the variants interface itself is a follow-up). The C# `internal` keyword is sidestepped via the verbatim identifier `@internal` for the wrapped instance field.

## Pipeline config

```yaml
output:
  languages:
    - csharp:
        namespace_root: 'Grafana.Foundation'   # default
        generate_json_converters: true
        generate_equality: false               # not yet implemented
        skip_runtime: false
```

The C# output is wired into `config/foundation_sdk.tests.yaml` so `make tests` / `gen-tests` exercises the full pipeline (16 generated `.cs` files committed under `testdata/generated/src/`).

## Implementation notes

- **Layout**: `internal/jennies/csharp/` mirrors `internal/jennies/java/`. Templates live under `templates/{types,marshalling,builders,runtime}/`.
- **Type formatter**: pre-formats every field/arg type in Go before rendering so the import map is fully populated when `{{ .Imports }}` is rendered. This is a deliberate deviation from Java (where `formatType` is called at render time and imports may be stale).
- **Cog references**: written as `Cog.X` and resolved via parent-namespace lookup — no `using Grafana.Foundation.Cog;` is emitted.
- **Source layout**: `src/Grafana/Foundation/<Package>/<Type>.cs` with a single `.csproj` for v1; per-package projects can be introduced later without touching paths or namespaces.

Each phase landed as a separate commit on this branch:

1. `371db9ad` — skeleton & wiring (`Language` interface, `OutputLanguage.CSharp`, pipeline switch)
2. `8326264d` — raw types jenny (89 golden files across 28 fixtures)
3. `3fa01876` — System.Text.Json marshalling
4. `909db276` — fluent builders + `Cog.IBuilder<T>` runtime (28 builder fixtures + smoke test)
5. `b092023d` — SDK test pipeline wiring + docs

## What's *not* in this PR (deferred)

- **Option converters** (Java-style object-to-builder-source codegen, used for "dashboards as code" import). Disabled by default in Java too; will be a follow-up.
- **`Equals` / `GetHashCode` jenny** behind `generate_equality`.
- **`Cog.Variants.<X>` interface generation** — currently referenced but not emitted.
- **Generic-panel hierarchy** (Java's `<T extends FooBuilder<T>>` for `Dashboard.Panel` / `VizConfigKind` / `DataQueryKind`).
- **Foundation-SDK sibling-repo wiring** (`examples/csharp`, `prepare-release.sh` hooks). `scripts/ci/build-csharp.sh` is added as a placeholder.

## Testing

- `go test ./...` is green.
- `internal/jennies/csharp/{rawtypes,converters,builder,runtime}_test.go` cover all jennies with golden files under `testdata/jennies/{rawtypes,builders,serializers,deserializers}/`.
- End-to-end `make gen-tests` produces a clean `testdata/generated/src/` tree (committed).
- The generated C# has not been compiled with `dotnet build` in this repo (no .NET toolchain in CI yet); spot-check by hand looks good. A real `csproj` + `dotnet build` step belongs in the foundation-sdk CI alongside `scripts/ci/build-csharp.sh`.

### DIFF
--- a/config/foundation_sdk.tests.yaml
+++ b/config/foundation_sdk.tests.yaml
@@ -7,6 +7,7 @@ parameters:
   go_package_root: 'github.com/grafana/cog/testdata/generated'
   php_namespace_root: 'Grafana\Foundation'
   java_package_path: 'com.grafana.foundation'
+  csharp_namespace_root: 'Grafana.Foundation'
 
 inputs:
   - cue:
@@ -28,3 +29,6 @@ output:
         generate_strict_unmarshaller: true
         generate_equal: true
         generate_validate: true
+    - csharp:
+        namespace_root: '%csharp_namespace_root%'
+        generate_json_converters: true
--- a/docs/pipelines/creating_pipeline.md
+++ b/docs/pipelines/creating_pipeline.md
@@ -307,6 +307,47 @@ output:
           - '%__config_dir%/templates/go/overrides'
 ```
 
+### C#
+
+```yaml
+output:
+  # …
+
+  languages:
+    - csharp:
+        # Root namespace for generated types. Generated source files land
+        # under `src/<namespace-as-folders>/<Package>/<Type>.cs`.
+        # Default: 'Grafana.Foundation'
+        namespace_root: 'Org.Package'
+
+        # Disables runtime-related code generation (e.g. `Cog.IBuilder<T>`).
+        # Note: builders can NOT be generated with this flag turned on, as
+        # they rely on the runtime to function.
+        # Default: false
+        skip_runtime: false
+
+        # Controls the generation of `System.Text.Json`-style
+        # `JsonConverter<T>` classes alongside generated types. Required
+        # for disjunction/discriminator support.
+        # Default: false
+        generate_json_converters: true
+
+        # Controls the generation of `Equals()` and `GetHashCode()`
+        # overrides on types.
+        # Default: false
+        generate_equality: false
+
+        # List of directories containing templates describing files to be
+        # added to the generated output.
+        extra_files_templates:
+          - '%__config_dir%/templates/csharp/extra'
+
+        # List of directories containing templates defining blocks used to
+        # override parts of builders/types/....
+        overrides_templates:
+          - '%__config_dir%/templates/csharp/overrides'
+```
+
 ### Java
 
 ```yaml
--- a/internal/codegen/output.go
+++ b/internal/codegen/output.go
@@ -1,6 +1,7 @@
 package codegen
 
 import (
+	"github.com/grafana/cog/internal/jennies/csharp"
 	"github.com/grafana/cog/internal/jennies/golang"
 	"github.com/grafana/cog/internal/jennies/java"
 	"github.com/grafana/cog/internal/jennies/jsonschema"
@@ -61,6 +62,7 @@ func (output *Output) interpolateParameters(interpolator ParametersInterpolator)
 }
 
 type OutputLanguage struct {
+	CSharp     *csharp.Config     `yaml:"csharp"`
 	Go         *golang.Config     `yaml:"go"`
 	Java       *java.Config       `yaml:"java"`
 	JSONSchema *jsonschema.Config `yaml:"jsonschema"`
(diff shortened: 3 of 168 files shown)


fdrstrok:
Friendly follow-up on this one 👋

Would you prefer:
- proceeding with this PR as-is (with any tweaks you suggest),
- me breaking it into smaller parts,
- or closing/reworking it differently?

Happy to adjust — just want to align on the best next step for this repo.

K-Phoen:
Hello there!

Thanks for the PR, but while I'd love to have C# support in cog, my bandwidth is limited and I don't think now is a good time to add C# (and my C# knowledge is quite... dated :sweat_smile:)

konnta0:
We have interest in using Foundation SDK from .NET, but maintaining a long-lived fork of both Cog and Foundation SDK would be difficult for adopters. From the project's perspective, what would need to be in place for C# support to become maintainable and potentially acceptable upstream?

fdrstrok:
Thanks @konnta0, that's the right question, and good to know about the bandwidth @K-Phoen. The aim is for this to be less work for you, not more.

The way I see it, C# support is two halves, and the first is already done in this PR.

The codegen half (done here). It follows the Java jenny closely and emits .NET with nullable reference types: raw types, opt-in System.Text.Json converters (including disjunction handling that mirrors Java's couldBe), fluent builders, and a small Cog.IBuilder<T> runtime. Every jenny has golden-file tests, it's wired into the SDK test pipeline, and "go test ./..." is green. All of it sits under internal/jennies/csharp/ and is opt-in via pipeline config, so no existing target changes.

The maintainability half (what I'm signing up for):

1. Add a .NET SDK to devbox and wire scripts/ci/build-csharp.sh to dotnet build the generated output. Once that's a required check, a red build is the signal and nobody has to read C# to review a change, same as gradle build does for Java.

2. Do the grafana-foundation-sdk side: runtime package, examples/csharp, NuGet packaging, and the prepare-release.sh / run-examples.sh hooks so it rides the existing workflow.

3. Land the deferred parity items as small follow-ups: option converters, Equals/GetHashCode, Cog.Variants emission, and the generic-panel hierarchy.

4. Own it. A CODEOWNERS entry on internal/jennies/csharp/, keeping it in sync as the IR evolves, and triaging C# issues so they don't land on you. @konnta0, given your .NET use case, I'd be glad to have you co-maintain.

The branch is already split into reviewable commits (skeleton, raw types, marshalling, builders, pipeline wiring), so I can break it into smaller PRs if that's easier. And if you'd rather keep C# experimental and undocumented until the CI build and SDK wiring are proven, that's fine by me. It lets people like @konnta0 drop their fork without committing the project to anything early.

No rush on timing. Happy to start wherever lowers the risk most for you, whether that's the dotnet build check or splitting this up.

PR #1101, opened 2026-05-11
Defaults: Fix empty map and slices as defaults

spinillos:
It adds some fixes for empty maps or slices generation. It also adds some tests to see the result per language.


testdata/jennies/rawtypes/struct_with_map_and_slice_default/GoRawTypes/defaults/types_gen.go
```diff
@@ -0,0 +1,123 @@
...
+		if string(fields["extra"]) != "null" {
+			if err := json.Unmarshal(fields["extra"], &resource.Extra); err != nil {
+				errs = append(errs, cog.MakeBuildErrors("extra", err)...)
+			}
+		} else {errs = append(errs, cog.MakeBuildErrors("extra", errors.New("required field is null"))...)
+		
+		}
+		delete(fields, "extra")
+	
+	}
```
IfSentient, line 72: Should we also have ab else here to error if `fields["extra"]` is `nil`, given it's required?

testdata/jennies/rawtypes/struct_with_map_and_slice_default/GoRawTypesUndiscriminated/defaults/types_gen.go
```diff
@@ -0,0 +1,90 @@
...
+		if string(fields["extra"]) != "null" {
+			if err := json.Unmarshal(fields["extra"], &resource.Extra); err != nil {
+				errs = append(errs, cog.MakeBuildErrors("extra", err)...)
+			}
+		} else {errs = append(errs, cog.MakeBuildErrors("extra", errors.New("required field is null"))...)
+		
+		}
+		delete(fields, "extra")
+	
+	}
```
IfSentient, line 71: Same thing about erroring if `extra` is absent.

IfSentient (commented):
Couple of comments/questions about the behavior when a required field is absent, not `null`.

testdata/jennies/rawtypes/struct_with_map_and_slice_default/GoRawTypes/defaults/types_gen.go
spinillos: Currently it only adds the error when the field is required but the default is nil: https://github.com/grafana/cog/blob/main/internal/jennies/golang/templates/types/struct.strict.json_unmarshal.tmpl#L34.

But the default is set in the constructor, so I understand that the value isn't set in this case and it should fail anyway, right?

internal/jennies/golang/rawtypes.go
```diff
@@ -245,144 +245,188 @@ func (jenny RawTypes) defaultsForStruct(context languages.Context, objectRef ast
...
 		objectName = referredPkg + "." + objectName
 	}
 
 	buffer.WriteString(objectName + "{\n")
 
-	extraDefaults := map[string]any{}
-	if val, ok := maybeExtraDefaults.(map[string]any); ok {
-		extraDefaults = val
-	}
+	extraDefaults, _ := maybeExtraDefaults.(map[string]any)
```
amalavet: it's ok that this could be nil now?

internal/jennies/golang/rawtypes.go
```diff
@@ -245,144 +245,188 @@ func (jenny RawTypes) defaultsForStruct(context languages.Context, objectRef ast
...
 
-				actualDefault := jenny.maybeValueAsPointer(defaultValue, true, disjunctionBranch.Type)
+	return buffer.String()
+}
 
-				nonNullableRefType := field.Type.DeepCopy()
-				nonNullableRefType.Nullable = false
+func fieldNeedsExplicitDefault(field ast.StructField, resolvedFieldType ast.Type, extraDefaults map[string]any) bool {
+	return field.Type.Default != nil ||
+		extraDefaults[field.Name] != nil ||
```
amalavet, line 281: if extraDefaults is passed in as nil here, it'll panic

internal/jennies/golang/rawtypes.go
```diff
@@ -245,144 +245,188 @@ func (jenny RawTypes) defaultsForStruct(context languages.Context, objectRef ast
...
+}
 
-				defaultValue = jenny.typeFormatter.formatRef(nonNullableRefType, false) + "{\n"
-				defaultValue += fmt.Sprintf("\t%s: %s,\n", formatFieldName(disjunctionBranchName), actualDefault)
-				defaultValue += "}"
+// defaultValueForField computes the Go literal for a field's default. The
+// second return value signals the caller to abort writing further fields
+// (used when a constant ref resolves to something we can't render).
+func (jenny RawTypes) defaultValueForField(context languages.Context, field ast.StructField, resolvedFieldType ast.Type, extraDefaults map[string]any) (string, bool) {
+	if extraDefault, ok := extraDefaults[field.Name]; ok {
```
amalavet, line 293: same here, if extraDefaults is nil, panic on lookup

internal/jennies/golang/rawtypes.go
```diff
@@ -245,144 +245,188 @@ func (jenny RawTypes) defaultsForStruct(context languages.Context, objectRef ast
...
-			t := context.ResolveRefs(ast.NewRef(constRef.ReferredPkg, constRef.ReferredType))
+	nonNullableRefType := field.Type.DeepCopy()
+	nonNullableRefType.Nullable = false
 
-			if !t.IsEnum() && !t.IsScalar() {
-				break
-			}
+	defaultValue = jenny.typeFormatter.formatRef(nonNullableRefType, false) + "{\n"
+	defaultValue += fmt.Sprintf("\t%s: %s,\n", formatFieldName(disjunctionBranchName), actualDefault)
+	defaultValue += "}"
```
amalavet: can this be done all as a `fmt.Sprintf`?

internal/jennies/golang/rawtypes.go
```diff
@@ -245,144 +245,188 @@ func (jenny RawTypes) defaultsForStruct(context languages.Context, objectRef ast
...
-			}
+	return defaultValue
+}
 
-			referredPkg = jenny.packageMapper(constRef.ReferredPkg)
-			if referredPkg != "" {
-				defaultValue = referredPkg + "." + defaultValue
-			}
+func (jenny RawTypes) defaultForMapWithDefault(field ast.StructField, resolvedFieldType ast.Type) string {
+	if emptyMap, ok := field.Type.Default.(map[string]any); ok && len(emptyMap) == 0 {
```
amalavet, line 352: need a nil check on Type.Default?

internal/jennies/golang/rawtypes.go
```diff
@@ -245,144 +245,188 @@ func (jenny RawTypes) defaultsForStruct(context languages.Context, objectRef ast
...
+		return "map[" + jenny.typeFormatter.formatType(resolvedFieldType.Map.IndexType) + "]" + jenny.typeFormatter.formatType(resolvedFieldType.Map.ValueType) + "{}"
+	}
+	return formatScalar(field.Type.Default)
+}
+
+func (jenny RawTypes) defaultForArrayWithDefault(field ast.StructField, resolvedFieldType ast.Type) string {
+	if emptySlice, ok := field.Type.Default.([]any); ok && len(emptySlice) == 0 {
```
amalavet, line 359: need a nil check on Type.Default?

internal/jennies/golang/rawtypes.go
```diff
@@ -245,144 +245,188 @@ func (jenny RawTypes) defaultsForStruct(context languages.Context, objectRef ast
...
+		return "[]" + jenny.typeFormatter.formatType(resolvedFieldType.Array.ValueType) + "{}"
+	}
+	return formatScalar(field.Type.Default)
+}
+
+func (jenny RawTypes) defaultForStructRefWithDefault(context languages.Context, field ast.StructField, resolvedFieldType ast.Type) string {
+	defaultValue := jenny.defaultsForStruct(context, *field.Type.Ref, resolvedFieldType, field.Type.Default)
```
amalavet, line 366: need a nil check on all this?

internal/jennies/golang/rawtypes.go
```diff
@@ -245,144 +245,188 @@ func (jenny RawTypes) defaultsForStruct(context languages.Context, objectRef ast
...
+	if field.Type.Nullable {
+		defaultValue = "&" + defaultValue
+	}
+	return defaultValue
+}
+
+func (jenny RawTypes) defaultForStructRef(field ast.StructField) string {
+	defaultValue := "New" + formatObjectName(field.Type.Ref.ReferredType) + "()"
```
amalavet, line 374: I'm afraid of all the nil checks that could be required on these ast values. having a method on these like `field.GetReferredType()` which would be access the field safely with a `val, ok := field.GetReferredType()` seems like it would make this all a bit less fragile.

internal/jennies/golang/rawtypes.go
```diff
@@ -245,144 +245,188 @@ func (jenny RawTypes) defaultsForStruct(context languages.Context, objectRef ast
...
-			defaultValue = "[]" + jenny.typeFormatter.formatType(field.Type.Array.ValueType) + "{}"
-		} else if field.Type.IsMap() {
-			defaultValue = "map[" + jenny.typeFormatter.formatType(field.Type.Map.IndexType) + "]" + jenny.typeFormatter.formatType(field.Type.Map.ValueType) + "{}"
-		} else {
-			defaultValue = "\"unsupported default value case: this is likely a bug in cog\""
+	return defaultValue
+}
+
+func (jenny RawTypes) defaultForEnumRef(field ast.StructField, resolvedFieldType ast.Type) string {
+	memberName := resolvedFieldType.Enum.Values[0].Name
```
amalavet, line 388: len > 0 check before accessing first val

testdata/jennies/rawtypes/struct_with_map_and_slice_default/GoRawTypes/defaults/types_gen.go
```diff
@@ -0,0 +1,123 @@
...
+
+	for field := range fields {
+		errs = append(errs, cog.MakeBuildErrors("SomeStruct", fmt.Errorf("unexpected field '%s'", field))...)
+	}
+
+	if len(errs) == 0 {
+		return nil
+	}
+
+	return errs
```
amalavet: nit; if no errors are added to the list, will it be nil on it's own, if so you can probably just return the errors rather than the len check first

testdata/jennies/rawtypes/struct_with_map_and_slice_default/GoRawTypes/defaults/types_gen.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+
+// Equals tests the equality of two `SomeStruct` objects.
+func (resource SomeStruct) Equals(other SomeStruct) bool {
+
```
amalavet, line 84: remove newline
```suggestion
```

internal/jennies/golang/rawtypes.go
spinillos: It doesn't panic, it returns nil. https://go.dev/play/p/Af0mRch4SPr

internal/jennies/golang/rawtypes.go
spinillos: AFAIK it should work as the same way if its nil. I'll put back the previous code because its less confusing.

internal/jennies/golang/rawtypes.go
spinillos: Enums always has values. Its why I didn't put the check.

testdata/jennies/rawtypes/struct_with_map_and_slice_default/GoRawTypes/defaults/types_gen.go
spinillos: This coming from equals template, but the templates aren't formatted.

testdata/jennies/rawtypes/struct_with_map_and_slice_default/GoRawTypes/defaults/types_gen.go
spinillos: I just updated the template. You will see a lot of tests updated.

internal/jennies/golang/rawtypes.go
spinillos: We already check if its a reference before reach this function.

amalavet (approved):

Issue #1085, opened 2026-04-22
Add deprecation option in schema flag.


spinillos:
Add deprecation option in schema configuration to indicate that a resource is deprecated and optionally, a message, the alternative and EOD.

K-Phoen:
Done at object level. We could add a transformation to mark all objects in a schema deprecated if needed.

PR #1107, opened 2026-05-14
Schemas: Set deprecation notice in an object

spinillos:
Contributes: https://github.com/grafana/cog/issues/1085

It creates a new action to set deprecation message in an object. Its useful when we have different versions of an object. For example, we have legacy `dashboard`, `dashboardv2beta1` and `dashboardv2`. 

We can add some information of the future of the schema.


K-Phoen:
I updated this PR and made deprecation messages also show up for builders and within the generated docs.

ryantxu (approved):

PR #1124, opened 2026-06-01
Bump codejen to v0.0.4

K-Phoen:
To take advantage of grafana/codejen#36 and perform case-insensitive filename checks when generating files from jennies.

This will detect/prevent issues like https://github.com/grafana/grafana-foundation-sdk/issues/1249


K-Phoen:
Note: the failures in Foundation SDK related jobs are precisely because we now detect https://github.com/grafana/grafana-foundation-sdk/issues/1249 at codegen-time :partying_face:

charandas (approved):

PR #1129, opened 2026-06-04
QoL improvements

K-Phoen:
A couple of nice-to-have improvements that I needed when debugging some stuff:
* the first commit makes sure that the `cog inspect` command doesn't initialize builder-related transforms when builders aren't being inspected
* the second commit makes the SchemaSetEntrypoint fail explicitly when we're trying to set an entrypoint that doesn't exist


charandas (approved):
⛵

PR #1130, opened 2026-06-05
Don't skip builder generation for objects implementing a variant

K-Phoen:
The way cog/the foundation SDK deals with dataqueries was slightly broken for dataqueries defined as a disjunction of types.

See https://github.com/grafana/grafana-foundation-sdk/issues/1204

Fixing this issue requires that cog deals slightly differently with these types.
For languages that do support disjunctions/union types (such as Python, PHP, ...) : no changes.
For other languages, we need to make sure that only the disjunction will be considered a dataquery and that a builder will be generated for it.

Some codegen configuration changes are also required on the SDK's side, that's what https://github.com/grafana/grafana-foundation-sdk/pull/1275 does.

### DIFF
--- a/internal/ast/compiler/disjunction_propagate_variant.go
+++ b/internal/ast/compiler/disjunction_propagate_variant.go
@@ -1,0 +1,49 @@
+package compiler
+
+import (
+	"fmt"
+
+	"github.com/grafana/cog/internal/ast"
+)
+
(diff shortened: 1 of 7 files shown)


K-Phoen:
The SDK-related failures are expected, changes to the configuration are required (see https://github.com/grafana/grafana-foundation-sdk/pull/1275)

ryantxu (approved):

PR #1135, opened 2026-06-10
Add Rust language target

jamesainslie:
## What

Adds a Rust language target to cog, capable of generating a complete, idiomatic Rust version of the Grafana Foundation SDK.

This addresses the Rust support request in grafana/grafana-foundation-sdk#1208.

## How

The target mirrors the structure of the existing python/golang jennies:

- `internal/jennies/rust/rawtypes.go` + `types.go`: type emission. Structs are PascalCase with snake_case fields and `#[serde(rename)]` preserving wire keys; not-required fields are `Option<T>` with `skip_serializing_if` (collections stay bare `Vec`/`HashMap` with `is_empty` skips, mirroring Go's `omitempty`); enums use serde renames (numeric enums via `serde_repr`); disjunctions become `#[serde(untagged)]` enums; constant references pin their values through `Default` impls; recursive references are `Box`ed (including cycles through `Option`).
- `internal/jennies/rust/builder.go`: builder emission. Each object gets a `<Name>Builder { internal, errors }` with a chaining option API; constraints accumulate `cog::BuildError`s surfaced by `build() -> Result<T, Vec<BuildError>>`; delegation takes `impl cog::Builder<T>`; envelopes, nested-path initialization (`get_or_insert_with`), factories and composable-slot options are supported.
- `internal/jennies/rust/templates/runtime/`: the hand-written runtime shipped with every generated crate: the `Builder<T>` trait, `BuildError`, and a variant registry (object-safe `Dataquery` trait objects with discriminator-keyed decoders and a lossless `UnknownDataquery` fallback, mirroring the Go runtime's `UnmarshalDataquery`).
- `internal/jennies/rust/plugins.go`: emits `src/cog/plugins.rs` registering each variant's decoder.
- `internal/jennies/rust/module_init.go`: emits `Cargo.toml`, `src/lib.rs` and the module tree.
- `internal/jennies/rust/postprocessor.go`: runs `rustfmt` over generated files (skipped gracefully when unavailable), mirroring the Go target's gofmt postprocessor.

## Testing

- Developed TDD against the existing golden-file harness: every rawtypes and builder fixture is in scope except those other targets also skip (`intersections` and `anonymous_struct` matching python, `dashboard_panel` matching python, `builder_delegation_in_disjunction` matching golang).
- Generating the full Foundation SDK from the Grafana kind-registry produces a crate (113 files) that passes `cargo build`, `cargo clippy --all-targets -- -D warnings` and `cargo fmt --check`.
- As an end-to-end check, a real 25-panel production dashboard (timeseries/stat/logs panels, prometheus and loki dataqueries, template variables) was rebuilt with the generated builders and its serialized JSON matched the reference dashboard structurally.

A follow-up PR to grafana-foundation-sdk would add the `rust:` entry to its `.cog/config.yaml` to start publishing the Rust SDK.


K-Phoen:
Hello!

Thanks for the PR, but while I'd love to have Rust support in cog, my bandwidth (and rust knowledge) are limited and I don't think now is a good time to add Rust.

PR #1258, opened 2026-09-21
fix(ir): only use a field as discriminator when its values are distinct

gillesdemey:
Generated decoders switch on a single field to work out which branch of a disjunction they are looking at, and the field was chosen as the first one present in every branch holding a concrete string. Its values were never checked for being distinct.

Candidates are also sorted before one is picked. Ranging over the map directly meant that, where more than one field would do, the chosen one could vary between runs.

**Example**

The following CUE definition would fail to discriminate properly, since it would find `type` first and not consider `variant` to discriminate on. In this example `type` has repeated values while `variant` should be considered instead.

```
#EmailV1:     { type: "email", variant: "email/v1",       settings: { addresses: string } }
#EmailMimir1: { type: "email", variant: "email/v0mimir1", settings: { to?: string } }
#SlackV1:     { type: "slack", variant: "slack/v1",       settings: { recipient?: string } }
#SlackMimir1: { type: "slack", variant: "slack/v0mimir1", settings: { channel?: string } }
```


gillesdemey:
Blocker for grafana/grafana#133106 — that PR declares each contact point integration type and version as its own CUE branch, and without this fix codegen picks `type` as the discriminator.

K-Phoen (approved):

K-Phoen:
LGTM, thanks :)

